PR #2, opened 2016-07-04
Functional updates

thefloweringash:


lib/active_record_view_model/update_data.rb
```diff
@@ -83,6 +83,44 @@ def self.extract_reference_metadata(hash)
...
+      update_type = update_type.to_sym
+
+      unless hash.has_key?(ActiveRecordViewModel::VALUES_ATTRIBUTE)
+        raise ViewModel::DeserializationError.new("Collection update missing '#{ActiveRecordViewModel::VALUES_ATTRIBUTE}' field")
+      end
+
+      values = hash.delete(ActiveRecordViewModel::VALUES_ATTRIBUTE)
+      unless values.is_a?(Array)
+        raise ViewModel::DeserializationError.new("Collection update has invalid '#{ActiveRecordViewModel::VALUES_ATTRIBUTE}' field: must be an array of values")
+      end
```
chrisandreae: json schema?

lib/active_record_view_model/update_data.rb
```diff
@@ -203,11 +247,23 @@ def parse(hash_data, valid_reference_keys)
...
-              unless value.is_a?(Array)
-                raise ViewModel::DeserializationError.new("Could not parse non-array collection association")
-              end
 
-              associations[name] = value.map { |child_hash| parse_association.(child_hash) }
+              associations[name] =
+                case
+                when value.is_a?(Array)
+                  children = value.map { |child_hash| parse_association.(child_hash) }
+                  [{ :op => :replace, :values => children }]
```
chrisandreae: I think input hashes use string keys everywhere else

chrisandreae: replace is necessarily incompatible with any other updates: is it fair to consider it of the same essential type and have it in an array?

lib/active_record_view_model/update_data.rb
```diff
@@ -203,11 +247,23 @@ def parse(hash_data, valid_reference_keys)
...
+                when value.is_a?(Hash) && value[ActiveRecordViewModel::TYPE_ATTRIBUTE] == ActiveRecordViewModel::FUNCTIONAL_UPDATE_TYPE
```
chrisandreae: Match all Hashes with this and let the schema check for the type attribute?

lib/active_record_view_model/update_data.rb
```diff
@@ -203,11 +247,23 @@ def parse(hash_data, valid_reference_keys)
...
 
             if association_data.collection?
-              unless value.is_a?(Array)
-                raise ViewModel::DeserializationError.new("Could not parse non-array collection association")
-              end
 
-              associations[name] = value.map { |child_hash| parse_association.(child_hash) }
+              associations[name] =
+                case
+                when value.is_a?(Array)
```
chrisandreae:
```case value
when Array
...
when Hash
...
end
```

lib/active_record_view_model/update_operation.rb
```diff
@@ -357,7 +357,48 @@ def build_updates_for_collection_association(association_data, association_updat
         clear_association_cache(model, association_data.reflection)
       end
 
-      child_viewmodels = resolve_child_viewmodels(association_data, association_update_datas, previous_child_viewmodels, update_context)
+      datas_for_previous_child_viewmodels = ->() do
```
chrisandreae: ouch that name

lib/active_record_view_model/update_operation.rb
```diff
@@ -331,14 +331,14 @@ def build_update_for_single_association(association_data, association_update_dat
...
-    def build_updates_for_collection_association(association_data, association_update_datas, update_context)
+    def build_updates_for_collection_association(association_data, association_update_ops, update_context)
       model = self.viewmodel.model
 
       # reference back to this model, so we can set the link while updating the children
       parent_data = ParentData.new(association_data.reflection.inverse_of, viewmodel)
 
-      unless association_update_datas.is_a?(Array)
-        raise ViewModel::DeserializationError.new("Invalid hash data array for multiple association: '#{association_update_datas.inspect}'")
+      unless association_update_ops.is_a?(Array)
```
chrisandreae: This check doesn't seem especially necessary now that we've fully preprocessed the association into functional update data

lib/active_record_view_model/update_operation.rb
```diff
@@ -357,7 +357,48 @@ def build_updates_for_collection_association(association_data, association_updat
...
+      end
+
+      new_child_datas =
+        association_update_ops.inject(nil) do |new_child_datas, update_op|
+          case update_op[:op]
+          when :replace
+            update_op[:values]
+
+          when :append
+            new_child_datas ||= datas_for_previous_child_viewmodels.()
```
chrisandreae: to me this has the smell of "replace" still actually being a special case

lib/active_record_view_model/update_data.rb
```diff
@@ -1,29 +1,83 @@
...
+        }
+      VIEWMODEL_UPDATE = JsonSchema.parse!(viewmodel_update)
+
+      collection_update_action =
+        {
+          'type'        => 'object',
+          'description' => 'collection functional update action',
+          'properties'  => {
+            TYPE_ATTRIBUTE   => { 'enum' => [
+              'append', # Append payload
```
chrisandreae: Could get this from `FunctionalUpdate::Type` to avoid duplication

lib/active_record_view_model/update_data.rb
```diff
@@ -218,11 +280,25 @@ def parse(hash_data, valid_reference_keys)
...
+                case value
+                when Array
+                  children = value.map { |child_hash| parse_association.(child_hash) }
+                  CollectionUpdate.new(CollectionUpdate::Type::Replace, children)
+
+                when Hash
+                  UpdateData.verify_schema!(Schemas::COLLECTION_UPDATE, value)
+                  functional_updates = value[ACTIONS_ATTRIBUTE].map do |action|
+                    type   = FunctionalUpdate::Type.parse!(action[TYPE_ATTRIBUTE])
+                    values = action[VALUES_ATTRIBUTE].map(&parse_association)
```
chrisandreae: is it ok to have nested data/updates in a deleted member, or should it be effectively reference only? If ok to have data, what does it mean?

thefloweringash: It should be reference only IMHO. Which is how this implementation behaves when it comes to `build!`, but it's currently silent about it. Will add more input validation.

lib/active_record_view_model/update_operation.rb
```diff
@@ -357,7 +354,58 @@ def build_updates_for_collection_association(association_data, association_updat
...
+
+            when FunctionalUpdate::Type::Remove
+              removed_refs = Set.new(fupdate.values.map(&:viewmodel_reference))
+              child_datas.reject! { |child_data| removed_refs.include?(child_data.viewmodel_reference) }
+
+            when FunctionalUpdate::Type::Update
+              new_datas = fupdate.values.group_by(&:viewmodel_reference)
+
+              new_datas.each do |ref, updates|
+                unless updates.length == 1
```
chrisandreae: This check doesn't seem to be quite enough - what if the same element is present in separate updates (or delete/append). Better to check uniqueness of all `association_update.update_datas.map(&:viewmodel_reference)`?

chrisandreae:
👍

lib/active_record_view_model/update_data.rb
```diff
@@ -285,6 +287,19 @@ def parse(hash_data, valid_reference_keys)
                   UpdateData.verify_schema!(Schemas::COLLECTION_UPDATE, value)
                   functional_updates = value[ACTIONS_ATTRIBUTE].map do |action|
                     type   = FunctionalUpdate::Type.parse!(action[TYPE_ATTRIBUTE])
+                    values = action[VALUES_ATTRIBUTE]
+
+                    # There's no reasonable interpretation of a remove update that includes data.
+                    # Report it as soon as we detect it.
+                    if type == FunctionalUpdate::Type::Remove
+                      invalid_entries = values.reject { |h| reference_only_hash?(h) }
```
chrisandreae: what if it's polymorphic?

lib/active_record_view_model/update_operation.rb
```diff
@@ -360,11 +362,25 @@ def build_updates_for_collection_association(association_data, association_updat
           association_update.values
 
         when CollectionUpdate::Type::Functional
-          child_datas =
+          child_datas        =
```
chrisandreae: whitespace?

lib/active_record_view_model/update_operation.rb
```diff
@@ -360,11 +362,25 @@ def build_updates_for_collection_association(association_data, association_updat
...
 
+          # Each updated child must be unique
+          duplicate_children = association_update
+                                 .update_datas
+                                 .duplicates { |upd| upd.viewmodel_reference }
+          if duplicate_children.present?
+            formatted_invalid_ids = duplicate_children.keys.map do |vm_ref|
+              name = vm_ref.viewmodel_class.view_name
+              id = vm_ref.model_id
+              "'#{name}(#{id})'"
```
chrisandreae: Is there a better place for this pretty printing? on ViewModelReference?

lib/active_record_view_model/update_data.rb
chrisandreae: what if it's has_many_through?

thefloweringash: polymorphic: as you point out, that's not a supported has_many type
has_many_through: entirely different codepath, will make a seperate pull request for this after merging this.

lib/active_record_view_model/update_operation.rb
thefloweringash: It's the IDE aligning it with the next assignment statement. I think the IDE style is wrong in this case, reduced to one space.

PR #3, opened 2016-07-04
Replace manual update parsing with more succinct json_schema

thefloweringash:


thefloweringash:
When specifying an array for non-collection association, this produces an error message like:

> Error parsing viewmodel update: For '#', [] is not an object.

Which is not as nice as it could be.

lib/active_record_view_model/update_data.rb
```diff
@@ -22,6 +46,18 @@ def has_key?(name)
...
     end
 
+    def self.describe_json_schema_error(error)
+      "Error parsing #{error.schema.description}: #{error.message}"
+    end
+
+    def self.verify_schema!(schema, value)
+      valid, errors = schema.validate(value)
+      unless valid
+        an_error = errors.first
```
chrisandreae: Why only return the first error?

chrisandreae:
Are the errors just strings? Can we get something we can parse more easily to return clearer messages?

thefloweringash:
Errors have a [decent-specification][error-specification]. Didn't want to implement a renderer for each error if I didn't have to.

[error-specification]: https://github.com/brandur/json_schema/blob/master/docs/errors.md

lib/active_record_view_model/update_data.rb
thefloweringash: Can return a human only readable version of the errors, as discussed, but we should revisit this when we redo the errors for path information.

lib/active_record_view_model/update_data.rb
```diff
@@ -1,7 +1,31 @@
...
 class ActiveRecordViewModel
   class UpdateData
     attr_accessor :viewmodel_class, :id, :attributes, :associations, :referenced_associations
 
+    module Schemas
+      REFERENCE        = JsonSchema.parse!(
+        {
+          'type'                 => 'object',
+          'description'          => 'shared reference',
+          'properties'           => { '_ref' => { 'type' => 'string' } },
```
chrisandreae: `ARVM::ReferenceAttribute` rather than `'_ref'`?

lib/active_record_view_model/update_data.rb
```diff
@@ -22,6 +46,23 @@ def has_key?(name)
...
     end
 
+    def self.verify_schema!(schema, value)
+      valid, errors = schema.validate(value)
+      unless valid
+        format_error = ->(e) { "#{e.schema.description}: #{e.message}" }
+
+        message =
+          if errors.length == 1
+            "Error parsing #{format_error.(errors.first)}"
```
chrisandreae: "Error(s)"?

chrisandreae:
Looks good

PR #4, opened 2016-07-08
Fix deferred model loading scope

thefloweringash:
The worklist will sometimes load a model outside of the update tree. It
must eagerly include associations inferred by the updates to be
applied. Prior to this change it was using the eager includes for
serialization, which in some cases will be insufficient.


chrisandreae:
lgtm

PR #7, opened 2016-07-12
Show changes 

thefloweringash:
Reopened from #6, which got eaten when branches changed.

### DIFF
--- a/lib/active_record_view_model/update_data.rb
+++ b/lib/active_record_view_model/update_data.rb
@@ -182,49 +182,88 @@ class ActiveRecordViewModel
       self.new(viewmodel.class, viewmodel.id, false, {}, [])
     end
 
+    # Normalised handler for looking at the UpdateData for an association
+    def reduce_association(name, value, empty:, map:, inject:)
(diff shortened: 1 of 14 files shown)


lib/active_record_view_model/update_data.rb
```diff
@@ -171,49 +171,88 @@ def self.empty_update_for(viewmodel)
       self.new(viewmodel.class, viewmodel.id, {}, [])
     end
 
+    # Normalised handler for looking at the UpdateData for an association
+    def visit_association(name, value, empty:, map:, reduce:)
```
chrisandreae: Please rename these. `reduce_association`? `process_association`? Just not `visit_` please.

chrisandreae:
Could we win by returning the `updated_associations` to the caller in the `DeserializeContext` so that we don't have to expose the `UpdateData` internals to callers? Then the controllers could just create a `SerializeContext` based on the results accumulated in the `DeserializeContext`

chrisandreae:
otherwise lgtm

PR #10, opened 2016-07-13
Rename gem to iknow_view_models

chrisandreae:


thefloweringash:
For lack of a better name, LGTM.

PR #11, opened 2016-07-14
Skip parse in UpdateData.empty_update_for

thefloweringash:


thefloweringash:
This was replaced by 861ab2d243a068adb9b0495470eace6be4f9ff4f to allow clients to specify empty refs to objects, instead of it being a library internal thing used during functional updates.

Though the above commit doesn't have tests, which it should take from this commit.

chrisandreae:
Open a new PR to take these tests?

PR #14, opened 2016-08-23
Fix functional updates scrambling ordered collections

thefloweringash:
Functional updates express a transformation from from old_data ->
new_data. When the old_data participated in an ordered collection this
ordering was ignored, causing any functional update to seemingly
scramble the order.

Additionally the old_data loading code unnecessarily replicated
viewmodel association loading, which does include the correct ordering.

### DIFF
--- a/lib/active_record_view_model/update_operation.rb
+++ b/lib/active_record_view_model/update_operation.rb
@@ -339,11 +339,14 @@ class ActiveRecordViewModel
       parent_data = ParentData.new(association_data.direct_reflection.inverse_of, viewmodel)
 
       # load children already attached to this model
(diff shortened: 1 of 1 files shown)


lib/active_record_view_model.rb
```diff
@@ -482,8 +482,16 @@ def delete_associated(association_name, associated, deserialize_context: self.cl
     end
   end
 
-  def read_association(association_name)
-    association_data = self.class._association_data(association_name)
+  def read_association(association_spec)
```
thefloweringash: It's either this or passing through `association_name` into `build_*`. Which do you prefer?

PR #15, opened 2016-08-24
Improve exceptions

chrisandreae:
Identify the nodes that caused deserialization to fail and reference them in exception metadata.


lib/view_model/deserialization_error.rb
```diff
@@ -0,0 +1,55 @@
...
+    def initialize(msg, nodes = [])
+      super(msg)
+      @nodes = Array.wrap(nodes)
+    end
+
+    def http_status
+      400
+    end
+
+    def metadata
```
thefloweringash: Should this include the exception type? Is that communicated to the client in any way other than the http status? I can't see here how the client would distinguish "stale update", which it should definitely handle internally, vs "request format was completely garbage", which is a API consumer error and can't reasonably be expected to be handled.

test/unit/active_record_view_model/has_many_test.rb
```diff
@@ -695,7 +711,7 @@ def test_functional_update_update_failure
       ParentView.deserialize_from_view(update_view)
     end
 
-    assert_match(/Stale update/, ex.message)
+    assert_match(/Stale functional update/, ex.message)
```
thefloweringash: This exception needs to be accurately communicated to the client. I'd expect a test for that here. See also other comment.

thefloweringash:
I was surprised to see only `node`s for every exception. I think it's a reasonable idea, much like a "blame" field.

thefloweringash:
LGTM ± the specific representation of the errors.

PR #16, opened 2016-08-26
Support read only arvm attributes

chrisandreae:


thefloweringash:
Change looks good to me. Tests wouldn't hurt.

PR #20, opened 2016-09-13
Add `blame_reference` to viewmodel

chrisandreae:
When deserializing, if an error occurs within a viewmodel, what viewmodel
should be reported as to blame. Can be overridden for example when a
viewmodel is merged with its parent.


chrisandreae:
So once we've got viewmodels, this works nicely. I'm not sure what's best to do at the UpdateData step where the ViewModel isn't constructed yet.

chrisandreae:
After discussion, decided that errors at the UpdateData step should always be client programmer errors rather than end user errors, so it's not so important to attach them.

PR #21, opened 2016-09-23
Update pre_parse_x interface (breaking API change)

chrisandreae:
* pre_parse methods now modify hash in place
* pre_parse_x now is passed the hash and the (removed) value
  from at that key, and is responsible for replacing the mutated
  value.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'activesupport', '~> 5.0'
 gem 'activerecord',  '~> 5.0'
 
 # Fetch private dependencies from git
-gem 'acts_as_manual_list',          git: 'git@github.com:iknow/acts_as_manual_list'
-gem 'iknow_params',                 git: 'git@github.com:iknow/iknow_params'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -8,7 +8,7 @@ gem 'activesupport', '~> 5.0'
 gem 'activerecord',  '~> 5.0'
 
 # Fetch private dependencies from git
-gem 'acts_as_manual_list',          git: 'git@github.com:iknow/acts_as_manual_list'
-gem 'iknow_params',                 git: 'git@github.com:iknow/iknow_params'
-gem 'cerego_active_record_patches', git: 'git@github.com:iknow/cerego_active_record_patches'
-gem 'deep_preloader',               git: 'git@github.com:iknow/deep_preloader'
+gem 'acts_as_manual_list',          git: 'git@github.com:iknow/acts_as_manual_list',          branch: 'master'
```
thefloweringash: Motivation for adding explicit branches is missing. Consider seperate commit to explain.

PR #23, opened 2016-10-13
Parsing and ancestors

chrisandreae:


lib/view_model/deserialize_context.rb
```diff
@@ -1,15 +1,29 @@
 class ViewModel
   class DeserializeContext
-    attr_accessor :updated_associations, :parent_ref
+    attr_accessor :updated_associations, :parent_context, :parent_viewmodel
 
-    def initialize(parent_ref: nil)
-      self.parent_ref = parent_ref
+    def initialize(*)
```
thefloweringash: Is this constructor necessary?

lib/view_model/deserialize_context.rb
```diff
@@ -1,15 +1,29 @@
...
+
+    def for_child(parent_viewmodel)
       self.dup.tap do |copy|
-        copy.parent_ref = parent.to_reference
+        copy.parent_context   = self
+        copy.parent_viewmodel = parent_viewmodel
       end
     end
+
+    private :parent_context, :parent_viewmodel
```
thefloweringash: Awkward to leave the writers public. Is it worth making a method with a different name `initialize_descendant` to make it more obvious?

lib/view_model/deserialize_context.rb
chrisandreae: Yes, because subclasses pass up arguments that get ignore. Better argument is perhaps that they _shouldn't_.

lib/view_model/deserialize_context.rb
chrisandreae: yes.

lib/view_model/reference.rb
```diff
@@ -1,9 +1,27 @@
 class ViewModel
   # Key to identify a viewmodel with some kind of inherent ID (e.g. an ActiveRecordViewModel)
-  Reference = Struct.new(:viewmodel_class, :model_id) do
+  class Reference
+    attr_accessor :viewmodel_class, :model_id
```
thefloweringash: Do these need writers?

thefloweringash:
I think this is the first commit that requires ruby 2.3.

chrisandreae:
Now that we have given in to requiring Rails 5, I'm ok with that dependency.

PR #28, opened 2016-10-31
Fix edit checks for has many through

chrisandreae:
* Edit check parent if children changed.
* Return updates to be added by caller, for consistency with other builders.
* Construct previous viewmodels in advance.
* Add tests for edit check of parent.

### DIFF
--- a/test/helpers/arvm_test_models.rb
+++ b/test/helpers/arvm_test_models.rb
@@ -99,6 +99,7 @@ class ARVMBuilder
 
   def initialize(name, &block)
     @name = name.to_s.camelize
+    @no_viewmodel = false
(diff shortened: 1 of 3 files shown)


test/helpers/arvm_test_models.rb
```diff
@@ -99,6 +99,7 @@ class ARVMBuilder
 
   def initialize(name, &block)
     @name = name.to_s.camelize
+    @no_viewmodel = false
```
thefloweringash: Motivation for this change?

chrisandreae: Use of uninitialized instance variable spams warnings at runtime.

PR #32, opened 2016-11-01
Improve ViewModel access checks

chrisandreae:
Provide information on what edits have been attempted: pass in `changed_associations` and `deleted` to `editable?`
Allow setting the error to be thrown without overriding `editable!`/`visible!`

### DIFF
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -172,23 +172,32 @@ class ViewModel
     ViewModel.preload_for_serialization([self], serialize_context: serialize_context)
   end
 
+  attr_writer :access_check_error
+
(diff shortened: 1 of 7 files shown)


thefloweringash:
This isn't the first place I've seen that it's a little bit awkward to carry around an `association_data` *and* and `association_name`. Does it yet make sense to put the name in the data, and consider the data the one true representation?

chrisandreae:
> Does it yet make sense to put the name in the data, and consider the data the one true representation?

I think yes, but can you point out the other places where we're carrying around a `(name,data)` pair? I've been making the change, but not found them yet.

PR #39, opened 2016-12-13
Has many through functional updates

thefloweringash:


lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+            appended_targets = fupdate.values.map do |ref|
+              {
+                ref:           ref,
+                viewmodel_ref: update_context.resolve_reference(ref).viewmodel_reference
+              }
+            end
+
+            if fupdate.before || fupdate.after
+              moved_refs =
+                fupdate.values.map { |ref| update_context.resolve_reference(ref).viewmodel_reference }.to_set
```
chrisandreae: You've already extracted these - why recalculate them rather than use the ones you've got?

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+      when ReferencedCollectionUpdate::Replace
+        target_indirect_viewmodels =
+          association_update.values.map do |ref|
+            {
+              ref:           ref,
+              viewmodel_ref: update_context.resolve_reference(ref).viewmodel_reference,
+            }
+          end
+
+      when ReferencedCollectionUpdate::Functional
```
chrisandreae: Some of the below could really do with some high level comments motivating the particular manipulations.

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+
+            if fupdate.before || fupdate.after
+              moved_refs =
+                fupdate.values.map { |ref| update_context.resolve_reference(ref).viewmodel_reference }.to_set
+
+              target_indirect_viewmodels =
+                target_indirect_viewmodels.reject { |target| moved_refs.include?(target[:viewmodel_ref]) }
+
+              ref   = fupdate.before || fupdate.after
+              index = target_indirect_viewmodels.find_index { |target| target[:viewmodel_ref] == ref }
```
chrisandreae: Does order need to matter? Should I be able to have a functional update that says [ "append y after x", "append x after a" ]?

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+        target_indirect_viewmodels =
+          previous_through_viewmodels.map { |vm| { viewmodel_ref: viewmodel_reference_for_indirect_model.(vm) } }
+
+        functional_update_check_duplicates!(association_update, update_context)
+
+        association_update.values.each do |fupdate|
+          case fupdate
+          when FunctionalUpdate::Append
+            appended_targets = fupdate.values.map do |ref|
+              {
```
chrisandreae: might be nice to introduce and motivate this type - it's not going to be consumed for another hundred lines or so, and it's not immediately apparent what it's about.

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+
+          when FunctionalUpdate::Update
+            # Just need to connect these through, so the ref is anchored somewhere in the update tree.
+            # Uniquess of vm_refs has already been checked.=
+            ref_by_indirect_vm_ref =
+              fupdate.values.index_by { |ref| update_context.resolve_reference(ref).viewmodel_reference }
+
+            target_indirect_viewmodels.each do |target|
+              associated_ref = ref_by_indirect_vm_ref.delete(target[:viewmodel_ref])
+              target[:ref]   = associated_ref if associated_ref
```
chrisandreae: oh! is that how they get `ref`s? I see! Gosh!

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+            removed_refs = fupdate.values.to_set
+            target_indirect_viewmodels.reject! { |target| removed_refs.include? target[:viewmodel_ref] }
+
+          when FunctionalUpdate::Update
+            # Just need to connect these through, so the ref is anchored somewhere in the update tree.
+            # Uniquess of vm_refs has already been checked.=
+            ref_by_indirect_vm_ref =
+              fupdate.values.index_by { |ref| update_context.resolve_reference(ref).viewmodel_reference }
+
+            target_indirect_viewmodels.each do |target|
```
chrisandreae: So a `FU::Update` could also update something that was put into the `target_indirect_viewmodels` by a `FU::Append`? Or is this prevented by the reference uniqueness check at parse time?

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+      else
+        raise_deserialization_error("Unknown association_update type '#{association_update.class.name}'")
+      end
+
+      new_through_viewmodels = target_indirect_viewmodels.map do |target|
+        target_viewmodel_ref, ref  = target.values_at(:viewmodel_ref, :ref)
 
         existing_through_viewmodel =
-          previous_through_viewmodels_by_indirect_ref[target_reference].try(:shift) if target_reference
+          (previous_through_viewmodels_by_indirect_ref[target_viewmodel_ref]&.shift if target_viewmodel_ref)
```
chrisandreae: So yep, on a double-take, yes those parentheses are necessary to get the assignment to always happen, and ooh, that feels to me like a sign to not do it. That kind of infix `if` for expression value rather than statement side-effect is not a quick visual parse, at least for me.

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+          else
+            raise ArgumentError.new("Unknown functional update: '#{fupdate.class}'")
+          end
+        end
+
+      else
+        raise_deserialization_error("Unknown association_update type '#{association_update.class.name}'")
+      end
+
+      new_through_viewmodels = target_indirect_viewmodels.map do |target|
```
chrisandreae: through == indirect? if so, let's settle on one name.

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+        target_viewmodel_ref, ref  = target.values_at(:viewmodel_ref, :ref)
 
         existing_through_viewmodel =
-          previous_through_viewmodels_by_indirect_ref[target_reference].try(:shift) if target_reference
+          (previous_through_viewmodels_by_indirect_ref[target_viewmodel_ref]&.shift if target_viewmodel_ref)
+
+        unless ref || existing_through_viewmodel
```
chrisandreae: so are we carrying around these pairs so that we can either construct or match existing? Would it be possible to instead match greedily rather than construct these intermediates?

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
+          raise_deserialization_error('Existing through model was not resolved')
```
chrisandreae: Also when would we hit this error case?

lib/view_model/active_record/update_operation.rb
```diff
@@ -526,35 +540,121 @@ def build_updates_for_has_many_through(association_data, reference_strings, upda
...
                                           position_setter: set_position)
       end
 
-      new_through_updates = new_through_viewmodels.zip(reference_strings, positions).map do |viewmodel, ref, position|
+      new_through_updates = new_through_viewmodels.zip(positions).map do |new_viewmodel, position|
+        viewmodel, ref = new_viewmodel.values_at(:viewmodel, :ref)
+
         update_data = UpdateData.empty_update_for(viewmodel)
-        update_data.referenced_associations[indirect_reflection.name] = ref # TODO layering violation
+        update_data.referenced_associations[indirect_reflection.name] = ref if ref # TODO layering violation
```
chrisandreae: When do we not have a `ref` here? If it's a previous child that hasn't been updated?

lib/view_model/active_record/update_data.rb
```diff
@@ -55,14 +160,158 @@ def initialize(values)
...
-        values.flat_map(&:values)
+      def update_datas(referenced_updates)
+        values.reject { |action| action.is_a?(FunctionalUpdate::Remove) }.flat_map(&:values)
+      end
+    end
+
+    class Parser < UpdateParser
+      attr_reader :parse_association
+
+      def initialize(association_data, blame_reference, parse_association:)
```
thefloweringash: I was going to leave the recursion in `UpdateData#parse`, but extracting it doesn't hurt.

lib/view_model/active_record/update_operation.rb
```diff
@@ -356,6 +356,28 @@ def build_update_for_single_association(association_data, association_update_dat
       end
     end
 
+    def functional_update_contents(update, update_context)
+      refs = update
+               .update_datas(update_context.referenced_update_data)
```
thefloweringash: This one is a bit awkward. Note that we want to use the update datas inside `UpdateContext#build` for preloading, which happens before `build!`. `resolve_reference`, which returns update ops, which are constructed in build, cannot work before `build!` is complete. We could add a parallel `resolve_reference_update_data` (± naming), that works before `build!` and gives us similar safety, but I don't think we can use the existing resolution method.

lib/view_model/active_record/update_operation.rb
thefloweringash: I have an unrelated data structure that happens to contain the same values. Changed to digging them out instead of recomputing.

lib/view_model/active_record/update_operation.rb
thefloweringash: would you prefer
```
foo =
  if cond
    1
  end
```
or
```
foo = nil
if cond
  foo = 1
end
```
?

lib/view_model/active_record/update_operation.rb
```diff
@@ -426,7 +441,7 @@ def build_updates_for_collection_association(association_data, association_updat
               end
 
             when FunctionalUpdate::Remove
-              removed_refs = Set.new(fupdate.values.map(&:viewmodel_reference))
+              removed_refs = fupdate.values.to_set
```
thefloweringash: It used to be acceptable for `Remove` types to have `UpdateData`s, which was misleading at best, since they'd never actually be part of the update op tree, and were constrained not to have any actual updates. In shared land we're much more careful about orphaned updates, so this was necessary, and I think clearer.

lib/view_model/active_record/update_data.rb
```diff
@@ -90,71 +341,86 @@ module Schemas
...
+          BEFORE_ATTRIBUTE          => viewmodel_reference_only,
+          AFTER_ATTRIBUTE           => viewmodel_reference_only
+        },
+      }
+
+      fupdate_owned =
+        fupdate_base.(ViewModel::Schemas::VIEWMODEL_UPDATE_SCHEMA)
+
+      fupdate_shared           =
+        fupdate_base.({ 'oneOf' => [ViewModel::Schemas::VIEWMODEL_REFERENCE_SCHEMA,
```
thefloweringash: At some point the sheer verbosity becomes more of a barrier.

lib/view_model/active_record/update_data.rb
```diff
@@ -90,71 +341,86 @@ module Schemas
...
+                                    viewmodel_reference_only] })
```
thefloweringash: See comment about `oneOf` in the structure that's up one level. TL;DR: `oneOf` is death for error messages, and I'm probably a bad for using it at all.

lib/view_model/active_record/update_data.rb
```diff
@@ -55,14 +160,158 @@ def initialize(values)
...
+      def initialize(association_data, blame_reference, valid_reference_keys:)
+        super(association_data, blame_reference)
+        @valid_reference_keys = valid_reference_keys
+      end
+
+      def action_schema
+        UpdateData::Schemas::REFERENCED_COLLECTION_UPDATE
+      end
+
+      def ref_hash?(value)
```
thefloweringash: I always instinctively write code as helpers followed by integration, mostly out of habit. This goes directly against using the implementation as private space for help methods following a `private` block. Happy to try a different convention which fits better with the constraints of the language. 

I think the value of naming things is worth keeping this one in the private block.

lib/view_model/active_record/update_data.rb
```diff
@@ -55,14 +160,158 @@ def initialize(values)
...
+      def action_schema
+        UpdateData::Schemas::COLLECTION_UPDATE
+      end
+
+      def parse_update_values(values)
+        values.map(&parse_association)
+      end
+
+      def parse_update_action(action)
+        ViewModel::Schemas.verify_schema!(UpdateData::Schemas::UPDATE_ACTION, action)
```
thefloweringash: Just a general comment on this set of types: Was never meant to be a grand abstract class hierarchy. The public interface is two "unrelated" parsers that handle collections, and their internals is the existing algorithm just parameterised. Though you're correct that it could be tidier, and I've put more effort in.

lib/view_model/active_record/update_operation.rb
thefloweringash: I've always thought of functional updates as list of actions that will be interpreted in order operating on the target collection. Do we need to revise this?

thefloweringash:
I've polished the existing code, added comments, and renamed fields to make their usage more obvious, but haven't done any grand redesigns. PTAL, and if it's still incomprehensible let's talk more about grand redesigns.

lib/view_model/active_record/update_data.rb
chrisandreae: I accept that, but as it is, my current thoughts reading pretty much any part of the file is "what is this type and what does it do?". The answer is often a stack of several method calls away and not at all easy to find.

lib/view_model/active_record/update_operation.rb
chrisandreae: The former is the same problem: I disapprove in non-exceptional cases of using the return value of an if without an else, because it's a bit unexpected and can easily cause incorrect human parses.

I like either

```ruby
foo = nil
foo = 1 if cond
```

or even alternatively

```ruby
foo = 1 if cond
```

and relying on local variables being automatically `nil` when introduced into scope. I just prefer to not rely on the return value of `if` except in a small number of special-ish cases where there's no possible accidental mis-parse. For example, I don't mind at all (and actually quite like) your previous use of 

```ruby
foo
  .map { |upd| upd.viewmodel_reference if upd.id }
  .compact
```

lib/view_model/active_record/collections.rb
```diff
@@ -2,7 +2,7 @@ class ViewModel::ActiveRecord
   module Collections
     def self.count_by(enumerable)
       enumerable.each_with_object(Hash.new(0)) do |el, counts|
-        key         = yield(el)
+        key         = block_given? ? yield(el) : el
```
chrisandreae: I wouldn't be bothered by that. Define `duplicates` as `duplicates_by{|x| x}`?

lib/view_model/active_record/update_data.rb
```diff
@@ -55,14 +160,158 @@ def initialize(values)
...
+      def initialize(association_data, blame_reference, parse_association:)
+        super(association_data, blame_reference)
+        @parse_association = parse_association
+      end
+
+      def action_schema
+        UpdateData::Schemas::COLLECTION_UPDATE
+      end
+
+      def parse_update_values(values)
```
chrisandreae: Ah, I see. I didn't actually understand at the time, I thought it was inside one of the fupdate classes, shadowing the `values` member. Mistake, but still a new name would prevent that confusion.

lib/view_model/active_record/update_data.rb
```diff
@@ -24,45 +18,325 @@ def self.for_type(type)
...
 
-    class Append < self
+    class Append
       NAME = 'append'
       attr_accessor :before, :after
-      def self.schema
-        UpdateData::Schemas::APPEND_ACTION
+      attr_reader :contents
+      def initialize(contents)
+        @contents = contents
```
chrisandreae: What type are the contents?

lib/view_model/active_record/update_data.rb
```diff
@@ -24,45 +18,325 @@ def self.for_type(type)
...
+  end
 
-    class Update < self
-      NAME = 'update'
-      def self.schema
-        UpdateData::Schemas::UPDATE_ACTION
+  # Parser for collection updates. Collection updates have a regular
+  # structure, but vary based on the contents. Updates to a direct,
+  # owned collection recurse deeply and create a tree of UpdateDatas,
+  # while through, shared collections collected reference strings.
```
chrisandreae: accidentally a word?

lib/view_model/active_record/update_data.rb
```diff
@@ -90,71 +366,86 @@ module Schemas
...
+      collection_update = ->(base_schema) do
         {
           'type'                 => 'object',
           'description'          => 'collection functional update',
           'additionalProperties' => false,
           'required'             => [ViewModel::TYPE_ATTRIBUTE, ACTIONS_ATTRIBUTE],
           'properties'           => {
             ViewModel::TYPE_ATTRIBUTE => { 'enum' => [FUNCTIONAL_UPDATE_TYPE] },
-            ACTIONS_ATTRIBUTE         => { 'type' => 'array', 'items' => base_functional_update_schema }
+            ACTIONS_ATTRIBUTE => { 'type' => 'array', 'items' => base_schema }
```
chrisandreae: might still want to check at this level that the TYPE is `{ 'enum' => [update, append, remove]}`, and then as you say, leave the specific validation of each of those to the subschemas.

lib/view_model/active_record/update_data.rb
chrisandreae: oh, is it that they're UpdateDatas in the case of direct and ref strings in the case of shared?

thefloweringash: They're used by both types of collections, so either ref strings or update datas.

lib/view_model/active_record/update_data.rb
thefloweringash: I believe this type is checked currently in the `fupdate_base_schema`.

lib/view_model/active_record/update_data.rb
```diff
@@ -24,45 +18,325 @@ def self.for_type(type)
...
+      @valid_reference_keys = valid_reference_keys
+    end
+
+    def parse(value)
+      case value
+      when Array
+        new_replace(parse_contents(value))
+
+      when Hash
+        ViewModel::Schemas.verify_schema!(action_schema, value)
```
chrisandreae: this is really an action_s_ schema, not just a single action, right?

lib/view_model/active_record/update_data.rb
chrisandreae: Oh, it does already! Oops.

lib/view_model/active_record/update_data.rb
thefloweringash: It's probably a `functional_update_schema`, i.e., the whole update, which has a list of actions, and its type, etc.

lib/view_model/active_record/update_operation.rb
```diff
@@ -495,66 +510,188 @@ def build_updates_for_collection_association(association_data, association_updat
...
+        target_viewmodel_ref, ref  = target.values_at(:viewmodel_ref, :ref)
 
-        existing_through_viewmodel =
-          previous_through_viewmodels_by_indirect_ref[target_reference].try(:shift) if target_reference
+        existing_direct_viewmodel =
+          if target_viewmodel_ref
+            previous_direct_viewmodels_by_indirect_ref[target_viewmodel_ref]&.shift
+          end
+
+        unless ref || existing_direct_viewmodel
```
chrisandreae: What could cause this?

lib/view_model/active_record/update_operation.rb
```diff
@@ -495,66 +506,206 @@ def build_updates_for_collection_association(association_data, association_updat
       child_updates
     end
 
-    # TODO name isn't generic like all others; why not _collection_referenced_?
-    def build_updates_for_has_many_through(association_data, reference_strings, update_context)
+    def shared_functional_update_contents(update, update_context)
+      refs = update
+               .references
+               .map { |upd| update_context.resolve_reference(upd).viewmodel_reference }
```
chrisandreae: `upd` is a reference string?

chrisandreae:
If you like it, I like it.

PR #41, opened 2016-12-21
Contextual edit checks

chrisandreae:


lib/view_model.rb
```diff
@@ -33,6 +33,14 @@ def view_name=(name)
       @view_name = name
     end
 
+    def debug_name=(name)
```
chrisandreae: please comment these with their purpose

lib/view_model.rb
```diff
@@ -212,6 +227,10 @@ def visible!(context: self.class.new_serialize_context)
     end
   end
 
+  def context_for_update(deserialize_context:)
```
chrisandreae: Please explain this in a comment: what kind of update is the context used for, when might you override this?

lib/view_model/active_record/update_operation.rb
```diff
@@ -158,6 +163,15 @@ def run!(deserialize_context:)
         end
       end
 
+      if self.released_children.present?
+        debug "-> #{debug_name}: Checking released children permissions"
+        self.released_children.reject(&:claimed?).each do |released_child|
```
chrisandreae: But can't a different subtree claim these children, before the entire process finishes?

lib/view_model/active_record/update_context.rb
```diff
@@ -3,18 +3,31 @@
 # the mechanism by which it is applied to models.
 class ViewModel::ActiveRecord
   class UpdateContext
+    attr_reader :referenced_update_data # TODO: eeeep
+
     ReleaseEntry = Struct.new(:viewmodel, :association_data) do
```
chrisandreae: Use of this class has escaped this file. Do you think that's enough reason to give it its own file?

lib/view_model/deserialize_context.rb
```diff
@@ -5,6 +5,7 @@ class DeserializeContext
     private :parent_context, :parent_viewmodel
 
     def initialize(*)
+      @type_contexts = {}
```
chrisandreae: Naming is hard. But this name is so abstract it could mean almost anything.

lib/view_model.rb
```diff
@@ -219,6 +219,12 @@ def visible!(context: self.class.new_serialize_context)
     end
   end
 
+  def begin_update(deserialize_context:)
```
chrisandreae: "`# Give models a chance to interact with the context`"

PR #42, opened 2016-12-21
Delete check in run

thefloweringash:


chrisandreae:
by the way, what was the point of `debug_name`? When's it different from `view_name`?

PR #43, opened 2016-12-22
Reduce metaprogramming required for attribute serializers/deserializers

chrisandreae:
Simplify attribute serializers/deserializers to make it easier to extend on the other side of the app/library boundary.


chrisandreae:
Todo here: drop `acts_as_enum` from VM::AR, update tests.

lib/view_model.rb
```diff
@@ -49,6 +53,15 @@ def attribute(attr, **args)
       _attributes << attr
     end
 
+    # An abstract viewmodel may want to define attributes to be shared by their
+    # subclasses. Redefine `_attributes` to close over the current class's
+    # _attributes and ignore children.
+    def lock_attribute_inheritance
+      _attributes.tap do |attrs|
+        define_singleton_method(:_attributes) { attrs }
```
thefloweringash: As discussed, since the `_attributes` return value is shared among subclasses, the result should be immutable.

diff --git a/lib/view_model.rb b/lib/view_model.rb
index 6a3b87e..0433c33 100644
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -59,6 +59,7 @@ class ViewModel
     def lock_attribute_inheritance
       _attributes.tap do |attrs|
         define_singleton_method(:_attributes) { attrs }
+        attrs.freeze
       end
     end

chrisandreae: done

PR #46, opened 2017-01-10
Rename `_` prefix on viewmodel meta-attributes to `$`

chrisandreae:
To resolve conflict with ElasticSearch metadata.

### DIFF
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -4,11 +4,11 @@ require 'jbuilder'
 require 'deep_preloader'
 
 class ViewModel
-  REFERENCE_ATTRIBUTE = "_ref"
+  REFERENCE_ATTRIBUTE = "$ref"
   ID_ATTRIBUTE        = "id"
-  TYPE_ATTRIBUTE      = "_type"
-  VERSION_ATTRIBUTE   = "_version"
-  NEW_ATTRIBUTE       = "_new"
+  TYPE_ATTRIBUTE      = "$type"
+  VERSION_ATTRIBUTE   = "$version"
+  NEW_ATTRIBUTE       = "$new"
 
   class << self
     attr_accessor :_attributes
--- a/lib/view_model/active_record.rb
+++ b/lib/view_model/active_record.rb
@@ -12,7 +12,7 @@ class ViewModel::ActiveRecord < ViewModel::Record
   # Defined before requiring components so components can refer to them at parse time
 
   # for functional updates
-  FUNCTIONAL_UPDATE_TYPE = "_update"
+  FUNCTIONAL_UPDATE_TYPE = "$update"
   ACTIONS_ATTRIBUTE      = "actions"
   VALUES_ATTRIBUTE       = "values"
   BEFORE_ATTRIBUTE       = "before"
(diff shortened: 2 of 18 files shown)


thefloweringash:
@chrisandreae is this PR in a state to be reviewed?

chrisandreae:
now that it's properly rebased, yes. Note that it's against `nested-routes-has-many-through`

chrisandreae:
Should we still do this, given we're not planning on (immediately) storing viewmodels directly in ElasticSearch? If we want to keep the door open to do that in the future, we should make this change before creating any mobile clients.

chrisandreae:
I couldn't decide before. It's a very small change, just requires some time to shepherd it through the release process on all three projects. However, as we're using ES now, it's entirely unnecessary. But if we don't do it, release a mobile client, and then decide we need to put viewmodels into elasticsearch in the future (for example if we give up on using Chewy), then we've locked ourselves in in a really nasty way.

thefloweringash:
How sad would you be if both ends implemented request middleware that distinguished `$` and `_` modes? 

Starting state: server writes with `_`, client writes with `_`.
1. Make server and client accept either
2. Wait an update period
3. Release a server and client that writes as `$`
4. Wait an update period
5. Drop support for `_` in both
6. ???
7. Profit.

It's a pain, but the lock-in is smaller if we're happy to assume that we have well defined update periods.

chrisandreae:
I think what makes me more sad is that despite having a web app with near instant deployments, and live editing, so there should never be more than a moment or so of client side unsaved data, we've managed to get into a mobile-like situation where users have long running copies of old code that we feel we have to support and few tools to ensure upgrades (I.e refresh) occur.

chrisandreae:
Thinking a bit more about it, it's also that there's a time balance. Will end users get disrupted? Only editors? Is it worth a day or so of dev time to save maybe 30 seconds of reload frustration each for five editors? Would any of the middleware scaffolding be useful for other purposes later, or would it only contribute to this specific change?

thefloweringash:
> Thinking a bit more about it, it's also that there's a time balance. Will end users get disrupted? Only editors? Is it worth a day or so of dev time to save maybe 30 seconds of reload frustration each for five editors? Would any of the middleware scaffolding be useful for other purposes later, or would it only contribute to this specific change?

Was only proposing a migration for this particular change that had the least inconvenience for users. If it's only editors, then sure, let's just cause some disruption and apologise. I'm always thinking a little of the long term though, in which we have end users using our app and don't want them to see error messages, _even if_ reloading is the solution.

chrisandreae:
Yeah, that long term scares me, which is in part why I'm pushing for adding as much clever handling and graceful fallback and automatic reloading when there aren't any local changes (or attempt to persist local changes, reload, then extract) and so on, just so that we can safely push changes without the end user experience being "oh shit, I saw one of those huge pink error screens"

chrisandreae:
but if we're going to spend any days on making migrations nicer, let's spend it on things in the above, rather than hacks to smooth over specific individual changes

chrisandreae:
We came up with a nicer way to build ES representations without Chewy; we're now unlikely to ever want to store full views there.

PR #49, opened 2017-01-23
Clean up ruby linter warnings.

chrisandreae:


lib/view_model.rb
```diff
@@ -42,7 +42,7 @@ def debug_name=(name)
     end
 
     def debug_name
-      @debug_name || view_name
+      instance_variable_defined?(:@debug_name) ? @debug_name : view_name
```
thefloweringash: As discussed, could do this as `@debug_name = nil` in the constructor, and use a definite `nil` as the absence of value instead of `undefined`. There's no need to distinguish which absence of debug name we have, so pick the one that's easier to read.

PR #51, opened 2017-01-30
Split editable check into editable/valid_edit.

chrisandreae:
This allows us to separate the check that a model (in its current state)
can be edited from the check that the specified changes may be made.
Particularly this allows us to delegate basic editability from a parent
view to children.

### DIFF
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -248,7 +248,7 @@ class ViewModel
   # and if found false, an error must be raised if an edit is later attempted.
   # To be overridden by viewmodel implementations.
   def editable?(deserialize_context: self.class.new_deserialize_context)
-    visible?(context: deserialize_context)
+    true
   end
 
   # Check that the attempted changes to this record are permitted in the given
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -232,36 +232,113 @@ class ViewModel
     @access_check_error.present?
   end
 
+  # Check that the user is permitted to view the record in its current state, in
+  # the given context. To be overridden by viewmodel implementation.
   def visible?(context: self.class.new_serialize_context)
     true
   end
 
-  def visible!(context: self.class.new_serialize_context)
-    self.access_check_error = nil
-    unless visible?(context: context)
-      raise case
-            when @access_check_error
-              @access_check_error
-            when context.is_a?(DeserializeContext)
-              DeserializationError::Permissions.new("Attempt to deserialize into forbidden viewmodel '#{self.class.view_name}'",
-                                                    self.blame_reference)
-            else
-              SerializationError::Permissions.new("Attempt to serialize forbidden viewmodel '#{self.class.view_name}'")
-            end
-    end
-  end
+  # Editable checks during deserialization are always a combination of
+  # `editable?` and `valid_edit?`, which express the following separate
+  # properties:
 
-  def editable?(deserialize_context: self.class.new_deserialize_context, changed_attributes:, changed_associations:, deleted:)
+  # Check that the record is eligible to be changed in its current state, in the
+  # given context. During deserialization, this must be called before any edits
+  # have taken place (thus checking against the initial state of the viewmodel),
+  # and if found false, an error must be raised if an edit is later attempted.
+  # To be overridden by viewmodel implementations.
+  def editable?(deserialize_context: self.class.new_deserialize_context)
     visible?(context: deserialize_context)
   end
 
-  def editable!(deserialize_context: self.class.new_deserialize_context, changed_attributes: [], changed_associations: [], deleted: false)
+  # Check that the attempted changes to this record are permitted in the given
+  # context. During deserialization, this must be called once all edits have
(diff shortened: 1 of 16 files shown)


test/helpers/arvm_test_models.rb
```diff
@@ -30,19 +30,24 @@ class ViewModelBase < ViewModel::ActiveRecord
...
-      @visible_checks = []
+      @editable_checks   = []
+      @valid_edit_checks = []
+      @visible_checks    = []
     end
 
-    def log_edit_check(viewmodel)
-      edit_checks << viewmodel.to_reference
+    def log_editable_check(viewmodel)
+      editable_checks << viewmodel.to_reference
```
thefloweringash: Would be nice to see some tests that inspect `editable_checks`.

lib/view_model.rb
```diff
@@ -232,39 +232,100 @@ def has_access_check_error?
...
+      raise_access_check_error do
+        DeserializationError::Permissions.new(
+          "Attempt to edit forbidden viewmodel '#{self.class.view_name}'",
+          self.blame_reference)
+      end
+    end
   end
 
-  def editable!(deserialize_context: self.class.new_deserialize_context, changed_attributes: [], changed_associations: [], deleted: false)
+  def was_editable!
```
thefloweringash: Consider a comment along the lines of

```
Editable checks for an entity are always combination of the two properties:

  - editable?
  - valid_edit?

`editable?` is called before any of the changes are considered. Note
that a failure from this method is delayed until the entity is known
to be edited, but the result is available to children.

`valid_edit?` is called at the end when the changes to this entity are
known.

When called as part of an entity tree update the call chain looks like

  - save_editable!
    - editable?
  - <recurse>
    - children while recursing can use `was_editable?` from their ancestors
      so an owned entity can be editable if their parent is.
  - was_editable!
    - ensure result of `save_editable!` was false
  - valid_edit!
    - valid_edit?

When called outside of an entity tree update, for example, when being
deleted, the call chain is as follows:

  - editable!
    - editable?
  - valid_edit!
    - valid_edit?
```

lib/view_model.rb
```diff
@@ -232,39 +232,100 @@ def has_access_check_error?
...
+  def valid_edit?(deserialize_context: self.class.new_deserialize_context, changes:)
+    true
+  end
+
+  # During deserialization, returns true if the viewmodel was `editable?` before
+  # any changes were attempted. Takes an error_view on which to optionally set
+  # the access check error, which can be used when delegating to a parent view's
+  # editability.
+  def was_editable?(error_view:)
+    unless @initial_editable_state
```
thefloweringash: `@initial_editable_state` doesn't seem to be initialised or cleared. Will generate warnings (even though it's going to throw right after), but also leave a little bit of state left over. Also tidier to be cleared around the time of `was_editable!`.

lib/view_model.rb
```diff
@@ -276,6 +337,14 @@ def ==(other_view)
   def hash
     self.class._attributes.map { |attr| self.send(attr) }.hash
   end
+
+  private
+
+  def raise_access_check_error
```
thefloweringash: Sometimes I wish block parameters could be nicely named. This is one of those times. But yay overall at this function. How would you feel about moving the `raise` into the callers?

lib/view_model.rb
chrisandreae: viewmodel implementations don't share an initialize, so it can't be quite as clean as you wanted. But I'll prepare the error and clear after was_editable!.

lib/view_model.rb
chrisandreae: yes, but I think we got ruby-syntaxed.

lib/view_model.rb
chrisandreae: hopefully improved with instructions for implementors.

lib/view_model/active_record/update_operation.rb
```diff
@@ -122,7 +127,10 @@ def run!(deserialize_context:)
         # comparing #foo, #foo_was, #new_record?. Note that edit checks for
         # deletes are handled elsewhere.
         if model.changed? || associations_changed?
-          viewmodel.editable!(deserialize_context: deserialize_context, changed_attributes: model.changed, changed_associations: @changed_associations)
+          viewmodel.was_editable!
+
+          viewmodel.valid_edit!(deserialize_context: deserialize_context,
+                                changes: ViewModel::DeserializeContext::Changes.new(changed_attributes: model.changed, changed_associations: @changed_associations))
```
chrisandreae: such shorter

test/helpers/arvm_test_models.rb
chrisandreae: semi difficult: note that it's stashing the call to `editable?` whether or not its result was used.

chrisandreae:
An additional change that would be nice is to explicitly call `visible!` before `*editable!` in all other deserialization situations so that editability no longer needs to delegate to visibility. This would make it easier to implement in applications.

PR #52, opened 2017-02-08
Minor changes to support caching

chrisandreae:
* Shared references no longer share `prune`/`include` serialize context from their root.
* `prune`/`include` normalization is now no longer private, so that library users can compare normalized specifications.
* `VM::AR#find` now supports multiple records, like modern AR's find.
* `dependent_viewmodels`/`deep_schema_version` now have the option to exclude shared references.

### DIFF
--- a/lib/view_model/active_record.rb
+++ b/lib/view_model/active_record.rb
@@ -127,19 +127,31 @@ class ViewModel::ActiveRecord < ViewModel::Record
       assocs.each { |assoc| association(assoc, **args) }
     end
 
-    ## Load an instance of the viewmodel by id
-    def find(id, scope: nil, eager_include: true, serialize_context: new_serialize_context)
+    ## Load instances of the viewmodel by id(s)
+    def find(ids, scope: nil, eager_include: true, serialize_context: new_serialize_context)
       find_scope = self.model_class.all
       find_scope = find_scope.merge(scope) if scope
 
-      ref = ViewModel::Reference.new(self, id)
-      model = ViewModel::DeserializationError::NotFound.wrap_lookup(ref) do
-        find_scope.find(id)
+      find_all = ids.is_a?(Array)
+      ids = Array.wrap(ids)
+
+      models = find_scope.where(id: ids).to_a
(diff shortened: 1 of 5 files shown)


lib/view_model/active_record.rb
```diff
@@ -218,25 +218,26 @@ def eager_includes(serialize_context: new_serialize_context)
...
+          vm.dependent_viewmodels(seen, include_shared: include_shared)
         end
       end
 
       seen
     end
 
-    def deep_schema_version
-      @deep_schema_version ||=
+    def deep_schema_version(include_shared: true)
```
thefloweringash: If you're feeling enthusiastic having tests for this would be nice.

lib/view_model/active_record.rb
```diff
@@ -127,19 +127,31 @@ def associations(*assocs, **args)
...
+    ## Load instances of the viewmodel by id(s)
+    def find(ids, scope: nil, eager_include: true, serialize_context: new_serialize_context)
       find_scope = self.model_class.all
       find_scope = find_scope.merge(scope) if scope
 
-      ref = ViewModel::Reference.new(self, id)
-      model = ViewModel::DeserializationError::NotFound.wrap_lookup(ref) do
-        find_scope.find(id)
+      find_all = ids.is_a?(Array)
+      ids = Array.wrap(ids)
```
thefloweringash: The error message when ids contains a duplicate is potentially confusing.

lib/view_model/serialize_context.rb
```diff
@@ -52,12 +61,14 @@ def extract_referenced_views!
...
     def serialize_references(json)
+      reference_context = self.for_references
+
       seen = Set.new
       while seen.size != @references.size
         @references.each do |ref, value|
           if seen.add?(ref)
             json.set!(ref) do
-              ViewModel.serialize(value, json, serialize_context: self)
+              ViewModel.serialize(value, json, serialize_context: reference_context)
```
thefloweringash: Yay principles!

PR #54, opened 2017-02-14
Render viewmodels after commit

thefloweringash:
Prevents DoubleRenderError when the commit fails. Typically applications
will `rescue_from` with some kind of `render` based handler. If we've
rendered inside the transaction then transaction failures will always
double render.

### DIFF
--- a/lib/view_model/controller.rb
+++ b/lib/view_model/controller.rb
@@ -84,12 +84,6 @@ module ViewModel::Controller
   end
 
   def render_json_string(response, status:)
-    ## jbuilder prevents this from working
-    ##  - https://github.com/rails/jbuilder/issues/317
-    ##  - https://github.com/rails/rails/issues/23923
-
-    # render(json: response, status: status)
-
-    render(plain: response, status: status, content_type: 'application/json')
+    render(json: response, status: status)
   end
 end
--- a/lib/view_model/controller.rb
+++ b/lib/view_model/controller.rb
@@ -9,8 +9,13 @@ module ViewModel::Controller
     end
   end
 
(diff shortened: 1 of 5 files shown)


thefloweringash:
This approach moves rendering outside of the transaction and attempts to ensure that all database access is completed within the transaction via preloading. There's the possibility of a mistake here, or worse, a requirement to load from the database in serialization.

The other approach is to try to bypass the exception from rails. At top level I believe the renderers are rack based, the return value is nicely propagated, and render is just an adapter for this. Conceptually there's no problem replacing a response, practically, *shudder*.

chrisandreae:
Dropping the safety of transactions makes me feel kind of anxious. I take your point that getting Rails to respect a replacement render would require digging deeper into the implementation internals than we'd like.
How about serializing the viewmodels to JSON or Hash inside the transaction and rendering the result outside it?

chrisandreae:
Could provide some scaffolding for it in ViewModel::Controller too, so it's not quite so fiddly. Something like `intemediate = prerender_viewmodel(...); end; render_viewmodel(intermediate)`

thefloweringash:
Did scaffolding as described above. Currently pretending we have an intermediate type that isn't JSON, so we can change it if we need to.

lib/view_model/controller.rb
```diff
@@ -10,7 +10,12 @@ module ViewModel::Controller
   end
 
   def render_viewmodel(viewmodel, status: nil, serialize_context: viewmodel.class.try(:new_serialize_context))
```
chrisandreae: doesn't pass a block to `prerender_viewmodel`

PR #57, opened 2017-03-07
viewmodel changes should be recorded as strings

chrisandreae:
Fix this, and add a test.

### DIFF
--- a/lib/view_model/active_record/update_operation.rb
+++ b/lib/view_model/active_record/update_operation.rb
@@ -49,7 +49,7 @@ class ViewModel::ActiveRecord
     end
 
     def association_changed!(association_name)
-      @changed_associations << association_name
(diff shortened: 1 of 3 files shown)


lib/view_model/active_record/update_operation.rb
```diff
@@ -49,7 +49,7 @@ def built?
     end
 
     def association_changed!(association_name)
-      @changed_associations << association_name
+      @changed_associations << association_name.to_s
```
thefloweringash: Yay this fixes it, but did you find out where the discrepancy comes from? Since most of our code works, would it be clearer to use `raise unless string`?

thefloweringash:
As discussed, the particular storage of these is exposed again via `changes_contained_to?`, normalise that too?

chrisandreae:
Done as requested :)

PR #60, opened 2017-04-03
Independent access control

chrisandreae:
Access control implemented directly in viewmodels is not flexible enough for our needs, and also awkwardly binds together controller and view concerns. Here we attempt to pull them out into a separate access control checking discipline that belongs to the controller.

### DIFF
--- a/lib/view_model/access_control.rb
+++ b/lib/view_model/access_control.rb
@@ -1,0 +1,130 @@
+require 'view_model/access_control_error'
+
+## Defines an access control discipline for a given action against a viewmodel.
+##
+## Access control is based around three edit check hooks: visible, editable and
+## valid_edit. The visible determines whether a view can be seen. The editable
+## check determines whether a view in its current state is eligible to be
+## changed. The valid_edit change determines whether an attempted change is
+## permitted. Each edit check returns a pair of boolean success and optional
+## exception to raise.
+class ViewModel::AccessControl
+  Result = Struct.new(:permit, :error) do
+    def initialize(permit, error: nil)
+      raise ArgumentError.new("Successful AccessControl::Result may not have an error") if permit && error
+      super(permit, error)
+    end
+
+    alias :permit? :permit
+
+    # Merge this result with another access control result. Takes a block
+    # returning a result, and returns a combined result for both tests. Access
+    # is permitted if both results permit. Otherwise, access is denied with the
+    # error value of the first denying Result.
+    def merge(&block)
+      if permit?
+        yield
+      else
+        self
+      end
+    end
+  end
+
+  Result::PERMIT = Result.new(true).freeze
+  Result::DENY   = Result.new(false).freeze
+
+  # Check that the user is permitted to view the record in its current state, in
+  # the given context.
+  def visible_check(view, context:)
+    Result::DENY
+  end
+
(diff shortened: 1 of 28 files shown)


chrisandreae:
TODO: association manipulation methods don't appropriately use a nested de/serialization context from the context that's passed in for the root. This will prevent root permission delegation from working with them.

lib/view_model/access_control.rb
```diff
@@ -0,0 +1,126 @@
...
+        view.blame_reference)
+    end
+  end
+
+  private
+
+  def ineligible(view)
+    # ARVM synthetic views are considered part of their association and as such
+    # are not edit checked.
+    # TODO: Do we want to make this a more generic scheme so that any VM type
```
chrisandreae: I don't currently expect this to be exposed to the application. Plan for enumerated types visibility is define an AccessControl::Tree that specifies them as visible/immutable and include it as relevant as an `ancestor`.

lib/view_model/access_control.rb
```diff
@@ -0,0 +1,126 @@
...
+  # Check that the attempted changes to this record are permitted in the given
+  # context. During deserialization, this must be called once all edits have
+  # been attempted. To be overridden by viewmodel implementations.
+  def valid_edit_check(view, deserialize_context:, changes:)
+    Result::DENY
+  end
+
+  # Implementations of deserialization that will potentially make changes to the
+  # viewmodel must call this on the viewmodel to save the initial `editable_check`
+  # value and optional exception before attempting to apply their changes.
```
chrisandreae: oops

lib/view_model/access_control.rb
```diff
@@ -0,0 +1,126 @@
...
+  Result = Struct.new(:permit, :error) do
+    def initialize(permit, error: nil)
+      raise "type error" if permit && error
+      super(permit, error)
+    end
+
+    alias :permit? :permit
+
+    # Merge this result with another access control result. Takes a block
+    # returning a result, and returns a combined result for both tests.
```
chrisandreae: names are hard :(

lib/view_model/access_control/composed.rb
```diff
@@ -0,0 +1,200 @@
...
+  # final
+  def valid_edit_check(view, deserialize_context:, changes:)
+    env = EditEnv.new(view, deserialize_context, changes)
+    check_delegates(env, edit_valid_ifs, edit_valid_unlesses)
+  end
+
+  protected
+
+  def visible_ifs
+    Lazily.concat(@visible_ifs,         *ancestors.map(&:visible_ifs))
```
chrisandreae: 👍  I think, but noticed another issue: the parent's `always` ancestor shows up multiple times.
`["child tree1", "child always", "parent always", "parent tree1", "parent always"]` I could keep a set and avoid visiting twice, but now I'm allocating again.

chrisandreae: Which I did, but any other suggestions?

test/unit/view_model_test.rb
```diff
@@ -2,52 +2,81 @@
 Bundler.require
 
 require "minitest/autorun"
+require "view_model/access_control/open"
+
+class OpenSerializeContext < ViewModel::SerializeContext
```
chrisandreae: no

PR #62, opened 2017-04-17
Extract method for wrapping array-or-object input

chrisandreae:
It's an extremely common pattern to wrap array-or-object input in an array, operate on it, and then return in the same format as when called. Replace ad-hoc implementations of pattern with a utility method.

### DIFF
--- a/lib/view_model/active_record.rb
+++ b/lib/view_model/active_record.rb
@@ -127,31 +127,26 @@ class ViewModel::ActiveRecord < ViewModel::Record
     end
 
     ## Load instances of the viewmodel by id(s)
-    def find(ids, scope: nil, eager_include: true, serialize_context: new_serialize_context)
+    def find(id_or_ids, scope: nil, eager_include: true, serialize_context: new_serialize_context)
       find_scope = self.model_class.all
       find_scope = find_scope.merge(scope) if scope
 
-      find_all = ids.is_a?(Array)
-      ids = Array.wrap(ids)
+      ViewModel::Utils.wrap_one_or_many(id_or_ids) do |ids|
 
-      models = find_scope.where(id: ids).to_a
+        models = find_scope.where(id: ids).to_a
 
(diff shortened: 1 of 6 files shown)


lib/view_model/active_record.rb
```diff
@@ -165,27 +160,20 @@ def load(scope: nil, eager_include: true, serialize_context: new_serialize_conte
...
-          updated_viewmodels.first
+        ViewModel::Utils.wrap_one_or_many(subtree_hash_or_hashes) do |subtree_hashes|
+          root_update_data, referenced_update_data = UpdateData.parse_hashes(subtree_hashes, references)
+
+          # Provide information about what was updated
+          deserialize_context.updated_associations = root_update_data
+                                                       .map { |upd| upd.updated_associations }
+                                                       .inject({}) { |acc, assocs| acc.deep_merge(assocs) }
+
+          _updated_viewmodels =
```
thefloweringash: Do you favour an effect-less assignment as documentation of return value significance?

lib/view_model/utils.rb
```diff
@@ -0,0 +1,15 @@
+class ViewModel::Utils
+  class << self
+    def wrap_one_or_many(obj)
```
thefloweringash: Naming is hard: `transform_as_array`? `as_array`? `on_array`? `use_as_array`?

lib/view_model/utils.rb
```diff
@@ -0,0 +1,15 @@
...
+      return_array = obj.is_a?(Array)
+      results = yield(Array.wrap(obj))
+      return_array ? results : results.first
+    end
+
+    def map_one_or_many(obj)
```
thefloweringash: This name bothers me less. It's fairly descriptive of what's happening.

chrisandreae:
Names are hard. Let's rename it if we come up with something.

PR #64, opened 2017-04-25
VM::AR::Visitor: introduce `visit_shared`

thefloweringash:
Allows parameterisation of visitation of shared associations.


thefloweringash:
Because this is a fairly minor change, and is backwards compatible, going to merge as is, and get a review when someone is available.

PR #66, opened 2017-04-27
ViewModel::ActiveRecord::Controller#destroy now returns 204 No Content.

OxleyS:


OxleyS:
I had only changed `ViewModel::ActiveRecord::Controller#destroy`, but it seems there is also `ViewModel::ActiveRecord::SingularNestedController` and `ViewModel::ActiveRecord::CollectionNestedController`. Like `destroy`, they all returned a `'data': null` response, except for `ViewModel::ActiveRecord::CollectionNestedController#disassociate_all`, which returns a `'data': []` response instead. What is the rationale for this?

chrisandreae:
The routes you mention are for manipulating an association, and returned the new value of the association: null in the case of a single association and empty list in case of multiple.

chrisandreae:
What do we want to do about this one?

PR #67, opened 2017-05-22
Check deferred database constraints before completing a VM::AR deserialization

chrisandreae:
Postgres supports deferring constraints until commit time. This can be an issue for library code as it can allow library-related errors to escape the scope of the library call. In ViewModel::ActiveRecord, force deferred constraints to be checked before returning from deserialization.

### DIFF
--- a/lib/view_model/active_record/update_context.rb
+++ b/lib/view_model/active_record/update_context.rb
@@ -165,6 +165,10 @@ class ViewModel::ActiveRecord
 
       @release_pool.release_all!
 
+      if updated_viewmodels.present?
+        check_deferred_constraints!(updated_viewmodels.first.model.class)
+      end
+
       updated_viewmodels
     end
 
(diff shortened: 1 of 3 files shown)


lib/view_model/active_record/update_context.rb
```diff
@@ -337,5 +341,21 @@ def try_take_released_viewmodel(vm_ref)
     def release_viewmodel(viewmodel, association_data)
       @release_pool.release_to_pool(viewmodel, association_data)
     end
+
+    # Deferred database constraints may have been violated by changes during
```
thefloweringash: Minor quibble: comment is describing behaviour not contained in the following code. The call to check and its motivation lives in `UpdateOperation.run!`.

lib/view_model/active_record/update_context.rb
```diff
@@ -165,6 +165,10 @@ def run!(deserialize_context:)
 
       @release_pool.release_all!
 
+      if updated_viewmodels.present?
```
thefloweringash: Confirm my understanding: it's only necessary to run when some model is changed, even in the presence of deferred foreign key constraints, because there's no way to delete a record without changing or deleting the record the constraint applies to.

test/unit/view_model/active_record_test.rb
```diff
@@ -468,4 +468,52 @@ def test_deserialize_context
...
+      List.connection.execute("ALTER TABLE lists ADD CONSTRAINT unique_child UNIQUE (child_id) DEFERRABLE INITIALLY DEFERRED")
+    end
+
+    class SentinelError < RuntimeError; end
+
+    def test_deferred_constraint_violation
+      l1 = List.create!(child: List.new)
+      l2 = List.create!
+
+      assert_raises(SentinelError) do
```
thefloweringash: A comment on why this is necessary would be nice. Given that test code is run within a transaction, and the transaction isn't committed, the deferred constraint will only be checked if checked explicitly. It seems like the test is equivalent to 

```ruby
ex = assert_raises { alter_by_view!(...)  {...} }
assert_match(ex, ...)
```

lib/view_model/active_record/update_context.rb
chrisandreae: I believe that's the case. That said, updated_viewmodels should always include all of the roots, whether changes were made or not, so the `present?` check is really only guarding against an empty operation.

PR #69, opened 2017-05-24
Update appraisals and add to travis

thefloweringash:


kevingriffin:
I published new versions of the gems that were locking this to 5.0. I also updated the gem spec to allow 5.1, and removed the restriction on JBuilder. The tests now pass.

PR #87, opened 2018-02-08
Import viewmodel features from other projects

chrisandreae:
In particular, ViewModel caching

### DIFF
--- a/test/unit/view_model/active_record/cache_test.rb
+++ b/test/unit/view_model/active_record/cache_test.rb
@@ -113,47 +113,47 @@ class ViewModel::ActiveRecord
       extend ActiveSupport::Concern
       included do
         it 'returns the right serialization' do
-          (serialize_with_cache).must_equal(serialize_from_database)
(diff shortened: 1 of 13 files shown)


test/unit/view_model/active_record/cache_test.rb
```diff
@@ -0,0 +1,276 @@
...
+
+    def serialize_with_cache
+      parse_result(fetch_with_cache)
+    end
+
+    module BehavesLikeACache
+      extend ActiveSupport::Concern
+      included do
+        it 'returns the right serialization' do
+          (serialize_with_cache).must_equal(serialize_from_database)
```
thefloweringash: Stray brackets? This pattern occurs a few times, and I'm not familiar with it. What's the significance?

chrisandreae: I mechanically converted from `expect(x).to equal(y)`. Converting and dropping the brackets looked like it lost something to me; the emphasis on the expressions that were being compared vanished. So I left them in. It is weird though. Do you think it adds enough to keep?

PR #90, opened 2018-03-22
Use child contexts when dealing with associations in nested controllers

chrisandreae:
Logically we're dealing with children of a top level entity, so construct child contexts before serialization.


chrisandreae:
Yes. Like we discussed though, testing that sort of behaviour through controller integration testing is pretty awkward.

PR #91, opened 2018-03-26
Traversal dependent rooting

chrisandreae:
Step one of the changes we discussed; does this part seem sensible?

### DIFF
--- a/lib/view_model/access_control/tree.rb
+++ b/lib/view_model/access_control/tree.rb
@@ -39,7 +39,7 @@ class ViewModel::AccessControl::Tree < ViewModel::AccessControl
 
       self::AlwaysPolicy.include_from(ancestor::AlwaysPolicy)
       ancestor.view_policies.each do |view_name, ancestor_policy|
-        policy = find_or_create_policy(view_name, root: ancestor_policy.root?)
+        policy = find_or_create_policy(view_name)
         policy.include_from(ancestor_policy)
       end
     end
@@ -47,12 +47,12 @@ class ViewModel::AccessControl::Tree < ViewModel::AccessControl
     def add_to_env(field_name)
       @env_vars << field_name
       self::AlwaysPolicy.add_to_env(field_name)
-      view_policies.values.each { |p| p.add_to_env(field_name) }
+      view_policies.each_value { |p| p.add_to_env(field_name) }
     end
 
     # Definition language
-    def view(view_name, root: false, &block)
-      policy = find_or_create_policy(view_name, root: root)
+    def view(view_name, &block)
+      policy = find_or_create_policy(view_name)
       policy.instance_exec(&block)
     end
 
@@ -66,7 +66,7 @@ class ViewModel::AccessControl::Tree < ViewModel::AccessControl
       policy = Class.new(Node)
       # View names are not necessarily rails constants, but we want
       # `const_set` them so they show up in stack traces.
-      mangled_name = view_name.gsub(".", "_")
+      mangled_name = view_name.tr('.', '_')
       const_set(:"#{mangled_name}Policy", policy)
       view_policies[view_name] = policy
       policy.include_from(self::AlwaysPolicy)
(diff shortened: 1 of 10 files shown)


test/unit/view_model/active_record/cache_test.rb
```diff
@@ -147,8 +147,15 @@ module BehavesLikeACache
             value(view_name).must_be(:present?)
```
thefloweringash: If you're doing misc linter cleanups, apparently this file has unnecessary directives:
```
diff --git a/test/unit/view_model/active_record/cache_test.rb b/test/unit/view_model/active_record/cache_test.rb
index 85bf8a8..ec05e7d 100644
--- a/test/unit/view_model/active_record/cache_test.rb
+++ b/test/unit/view_model/active_record/cache_test.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-# rubocop:disable Style/RedundantParentheses, Lint/MissingCopEnableDirective
-
 require "minitest/autorun"
 require "minitest/unit"
 require "minitest/hooks"
```

test/unit/view_model/active_record/has_many_through_poly_test.rb
```diff
@@ -54,7 +54,7 @@ def self.build_parent(arvm_test_case)
 
       define_viewmodel do
         attributes :name
-        association :tags, shared: true, through: :parents_tags, through_order_attr: :position, viewmodels:[TagAView, TagBView]
+        association :tags, shared: true, optional: false, through: :parents_tags, through_order_attr: :position, viewmodels: [TagAView, TagBView]
```
thefloweringash: Any reason to specify the default here?

lib/view_model/access_control/tree.rb
```diff
@@ -169,53 +142,48 @@ def inherited(subclass)
...
+        root_children_editable_unlesses << new_permission_check(reason, &block)
+      end
+
+      def root?
+        @root
       end
 
       def inspect_checks
         checks = super
-        checks.unshift("root: #{root?}")
```
thefloweringash: Any reason for removing this? Alternately, consider:
```ruby
if root?
  checks << "no root checks"
else
  checks << "root_children_visible_if: ..."
  checks << "root_children_visible_unless: ..."
  checks << "root_children_editable_if: ..."
  checks << "root_children_editable_unless: ..."
end
```

lib/view_model/access_control/tree.rb
```diff
@@ -226,51 +194,55 @@ def initialize(tree_access_control)
...
-        end
+        root_data = deserialize_context.nearest_root.descendent_access_control_data
+        root_data.editability.merge { super }
       end
     end
 
     private
 
+    def validate_root!(view, context)
+      if self.class.root? && !context.root?
```
thefloweringash: Consider `self.class.requires_root?` for clarity.

test/unit/view_model/access_control_test.rb
```diff
@@ -259,15 +276,15 @@ def test_visibility_from_root
       end
 
       refute_serializes(Tree1View, make_tree("visible", "invisible"))
-      assert_serializes(Tree1View, make_tree("visible_children", "invisible"))
+      assert_serializes(Tree1View, make_tree("visible_children", "visible"))
```
thefloweringash: This change, while technically the same test, seems to be the opposite spirit of the test. The change from `visible` to `visible_children` should be sufficient to change the `refute` to an `assert`. This implies that more changes were necessary.

chrisandreae: I think this was just wrong before. It asserted it serialized, meaning everything should've been visible.

chrisandreae: (saying `invisible` didn't break it: it still passed, it was just super confusing)

thefloweringash: There are two ways of reading this:
 - all expressions are controlling the access control
 - some expressions controlling access control and some talking to the reader about what the expected result is

I was doing the first, you're doing the second. The following test also is opposite. Pick one?

chrisandreae: I didn't realise they weren't all doing the second. My preference is the second (or at least a variation: _all_ expressions tell the reader what their result is, some of the values are also meaningful to the access control in order to cause that result as well) but I could be convinced otherwise?

chrisandreae: How about these changes?

chrisandreae:
Closing in favour of #92

PR #92, opened 2018-04-11
Traversal callbacks

chrisandreae:
**Work in progress**
[Based on traversal-dependent-rooting branch]

Adds callbacks to viewmodel traversals (serialization/deserialization). Rewrites AccessControl as a callback discipline.

### DIFF
--- a/lib/view_model/access_control.rb
+++ b/lib/view_model/access_control.rb
@@ -68,6 +68,24 @@ class ViewModel::AccessControl
     Result::DENY
   end
 
+  # Wrappers to check access control for a single view directly. Because the
+  # checking is run directly on one node without any tree context, it's only
+  # valid to run:
+  # * on root views
+  # * when no children could contribute to the result
+  def visible!(view, context:)
+    run_callback(ViewModel::Callbacks::Hook::BeforeVisit, view, context)
+    run_callback(ViewModel::Callbacks::Hook::AfterVisit,  view, context)
+  end
+
+  def editable!(view, deserialize_context:, changes:)
+    run_callback(ViewModel::Callbacks::Hook::BeforeVisit,       view, context)
+    run_callback(ViewModel::Callbacks::Hook::BeforeDeserialize, view, context)
+    run_callback(ViewModel::Callbacks::Hook::OnChange,          view, context, changes: changes) if changes
+    run_callback(ViewModel::Callbacks::Hook::AfterDeserialize,  view, context, changes: changes)
+    run_callback(ViewModel::Callbacks::Hook::AfterVisit,        view, context)
+  end
+
   # Edit checks are invoked via traversal callbacks:
   include ViewModel::Callbacks
 
@@ -77,14 +95,9 @@ class ViewModel::AccessControl
     result = visible_check(self)
 
     raise_if_error!(result) do
-      message =
-        if context.is_a?(ViewModel::DeserializeContext)
-          "Attempt to deserialize into forbidden viewmodel '#{view.class.view_name}'"
-        else
-          "Attempt to serialize forbidden viewmodel '#{view.class.view_name}'"
-        end
-
-      ViewModel::AccessControlError.new(message, view.blame_reference)
+      ViewModel::AccessControlError.new(
+        "Illegal access to viewmodel '#{view.class.view_name}'",
+        view.blame_reference)
     end
   end
 
@@ -113,6 +126,7 @@ class ViewModel::AccessControl
 
   after_deserialize do
     next if ineligible(view)
+    # If there was no change to consume the initial editability we still want to clean it up
     cleanup_editability(view)
   end
 
--- a/lib/view_model/access_control.rb
+++ b/lib/view_model/access_control.rb
@@ -41,7 +41,7 @@ class ViewModel::AccessControl
 
   # Check that the user is permitted to view the record in its current state, in
   # the given context.
-  def visible_check(_view, context:)
+  def visible_check(_traversal_env)
     Result::DENY
   end
 
@@ -55,7 +55,7 @@ class ViewModel::AccessControl
   # checking against the initial state of the viewmodel), and if editing is
   # denied, an error must be raised only if an edit is later attempted. To be
   # overridden by viewmodel implementations.
-  def editable_check(_view, deserialize_context:)
+  def editable_check(_traversal_env)
     Result::DENY
   end
 
@@ -64,17 +64,17 @@ class ViewModel::AccessControl
   # transactional backing models, the changes may be made in advance to give the
   # edit checks the opportunity to compare values. To be overridden by viewmodel
   # implementations.
-  def valid_edit_check(_view, deserialize_context:, changes:)
+  def valid_edit_check(_traversal_env)
     Result::DENY
   end
 
   # Edit checks are invoked via traversal callbacks:
   include ViewModel::Callbacks
 
-  before_visit do |view, context:|
+  before_visit do
     next if ineligible(view)
 
-    result = visible_check(view, context: context)
+    result = visible_check(self)
 
     raise_if_error!(result) do
       message =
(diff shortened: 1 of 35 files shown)


lib/view_model/callbacks.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+
+      unless pos_params.size == 1
+        raise ArgumentError.new("Cannot add callback to hook #{hook}: "\
+                                "must have exactly one positional parameter.")
+      end
+
+      key_param_names = key_params.map { |_type, name| name }
+      unless (required_params.to_set ^ key_param_names).blank?
+        raise ArgumentError.new("Cannot add callback to hook #{hook}: "\
```
thefloweringash: This error looks misleading when you are missing a required parameter. Add `"expected: #{required_params}"` here.

lib/view_model/access_control.rb
```diff
@@ -89,17 +88,22 @@ def visible!(view, context:)
...
 
-    result = initial_editability.merge do
+    save_editability(view, initial_result)
+  end
+
+  on_change do |view, changes:, deserialize_context:|
+    next if ineligible(view)
+    initial_result = fetch_editability(view)
+
+    next if changes.nil?
```
thefloweringash: Will this case ever be hit?

lib/view_model/active_record/update_operation.rb
```diff
@@ -131,33 +129,36 @@ def run!(deserialize_context:)
...
-          deserialize_context.editable!(viewmodel,
-                                        initial_editability: initial_editability,
-                                        changes: changes)
+          # The hooks before this might have caused additional changes. All
+          # changes should be checked by the policy, so we have to recalculate
+          # the change set. Note that changes to associated children that point
+          # to this parent are recorded in `changes`, but their models have not
+          # yet been updated.
+          final_changes = viewmodel.changes
+          deserialize_context.run_callback(ViewModel::Callbacks::Hook::OnChange, viewmodel, changes: final_changes)
```
thefloweringash: Can we `raise` if `changes` changed here?

lib/view_model/active_record/update_operation.rb
```diff
@@ -131,33 +129,36 @@ def run!(deserialize_context:)
...
-
-        debug "-> #{debug_name}: Saving"
-        begin
-          model.save!
-        rescue ::ActiveRecord::RecordInvalid => ex
-          raise ViewModel::DeserializationError::Validation.from_active_model(ex.errors, blame_reference)
-        rescue ::ActiveRecord::StaleObjectError => ex
-          raise ViewModel::DeserializationError::LockFailure.new(blame_reference)
+        # Save if the model has been altered. Covers not only models with
+        # view changes but also lock version assertions.
```
thefloweringash: If the lock version causes `changed?` then we already know we've failed. I think the code is fine, but the lock version assertions with no other changes seem to not be supported in rails, unless also using explicit `.lock()`.

lib/view_model/active_record/nested_controller_base.rb
```diff
@@ -8,14 +8,16 @@ def show_association(scope: nil, serialize_context: new_serialize_context)
...
-      serialize_context.visible!(owner_view)
+      serialize_context.run_callback(ViewModel::Callbacks::Hook::BeforeVisit, owner_view)
       # Association manipulation methods construct child contexts internally
       associated_views = owner_view.load_associated(association_name, scope: scope, serialize_context: serialize_context)
 
       associated_views = yield(associated_views) if block_given?
 
       child_context = serialize_context.for_child(owner_viewmodel, association_name: association_name)
-      prerender_viewmodel(associated_views, serialize_context: child_context)
+      json = prerender_viewmodel(associated_views, serialize_context: child_context)
```
thefloweringash: I had to go look at the implementation to see if this was actually json. Everywhere else leaves it opaque. Consider `.tap` or assigning to `pre_rendered` explicitly.

lib/view_model/active_record/association_manipulation.rb
```diff
@@ -200,29 +201,37 @@ def delete_associated(association_name, associated_id, type: nil, deserialize_co
...
-      child_context.visible!(vm)
-      initial_editability = child_context.initial_editability(vm)
-      child_context.editable!(vm,
-                              initial_editability: initial_editability,
-                              changes: ViewModel::Changes.new(deleted: true))
-      association.delete(vm.model)
-      vm
+      child_vm = direct_viewmodel.new(models.first)
+
+      child_context.run_callback(ViewModel::Callbacks::Hook::BeforeVisit, child_vm)
```
thefloweringash: This is quite a mouthful. We don't have any overlapping regions, so we could make some helpers that help balancing callbacks:

```end
def with_foo_callback(context, vm)
  context.run_callback(BeforeFoo, vm)
  yield
  context.run_callback(AfterFoo, vm)
end
```

Trading explicit calls for some indent levels

lib/view_model/access_control/tree.rb
```diff
@@ -109,51 +104,41 @@ def valid_edit_check(view, deserialize_context:, changes:)
...
+    if @root_editability_store.has_key?(node.object_id)
+      raise RuntimeError.new("Root access control data already saved for root")
     end
+    @root_editability_store[node.object_id] = descendent_editability
+  end
 
-    def initialize_as_child(*)
-      @access_control_root_data = nil
+  def fetch_descendent_editability(node)
+    # puts "fetching edit root for #{node.to_reference}" # TODO remove debug
```
thefloweringash: TODO (x4)

lib/view_model/access_control/tree.rb
```diff
@@ -109,51 +104,41 @@ def valid_edit_check(view, deserialize_context:, changes:)
...
-  # child-editability/visibility for tree-based access control roots.
-  module AccessControlRootMixin
-    extend ActiveSupport::Concern
-
-    RootData = Struct.new(:visibility, :editability)
-
-    def initialize(*)
-      super
-      @access_control_root_data = nil
+  def store_descendent_editability(node, descendent_editability)
```
thefloweringash: `node` parameter is of type `ViewModel`, not `Node`.

lib/view_model/callbacks.rb
```diff
@@ -0,0 +1,119 @@
...
+      @required_params = [context_name, *other_params]
+    end
+
+    def dsl_name
+      name.underscore
+    end
+  end
+
+  # Placeholder for callbacks to invoked for all view types
+  ALWAYS = "__always"
```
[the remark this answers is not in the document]
chrisandreae: Discussed out of band: at least in principle view names are constrained in the same way as Ruby type names, so this cannot collide as is. We might want to add explicit constraints for view names, since I think we don't have them yet.

lib/view_model/access_control.rb
chrisandreae: not any more, after introducing the onChange hook. deleted.

lib/view_model/active_record/update_operation.rb
chrisandreae: where?

lib/view_model/active_record/update_operation.rb
chrisandreae: Nothing to do here, but acknowledged that if `model.changes` includes `lock_version` we know already that it would fail due to lock assertion before going to the database. I don't think it's worth checking that explicitly though. Maybe PR to Rails to add that check in there instead.

lib/view_model/active_record/nested_controller_base.rb
chrisandreae: done

lib/view_model/active_record/association_manipulation.rb
chrisandreae: I think I prefer the explicit calls, given how big the body in the yield is. What do you think?

thefloweringash: Either seems fine.

lib/view_model/access_control/tree.rb
chrisandreae: fixed

lib/view_model/access_control/tree.rb
chrisandreae: fixed

lib/view_model/active_record/association_manipulation.rb
chrisandreae: Tried out the wrappers, on balance I think they're worth it.

PR #104, opened 2018-09-04
Allow the viewmodel cache to depend on an external group

thefloweringash:
When there are multiple views on a single model, they must mutually invalidate. Presently the cache is entirely internal and unique per viewmodel. The two approaches that come to mind are as follows.

### [A] All cachable models must provide the cache

Which means an extra restriction that whenever you include `ViewModel::ActiveRecord::Cache`, you must also define a cache getter on the model. Failing to define the cache getter will explode at load time. :+1: 

### [B] Let the user specify the cache when necessary (this PR)

Less disruptive and backwards compatible. Lets you use the defaults when you only have a single cache, and doesn't expose the details of the caching libray. But it seems a bit like a loaded foot gun. Any failure to override the method, like a simple typo, and the cache defaults to the internal cache.

### DIFF
--- a/lib/view_model/active_record/cache.rb
+++ b/lib/view_model/active_record/cache.rb
@@ -10,22 +10,11 @@ class ViewModel::ActiveRecord::Cache
 
   attr_reader :viewmodel_class
 
-  def initialize(viewmodel_class)
+  # If cache_group: is specified, it must be a group of a single key: `:id`
+  def initialize(viewmodel_class, cache_group: nil)
     @viewmodel_class = viewmodel_class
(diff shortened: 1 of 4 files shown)


lib/view_model/active_record/cache.rb
```diff
@@ -10,32 +10,22 @@ class UncacheableViewModelError < RuntimeError; end
 
   attr_reader :viewmodel_class
 
-  def initialize(viewmodel_class)
+  # If model_cache_group: is specified, it must be a group of a single key: `:id`
+  def initialize(viewmodel_class, model_cache_group: nil)
```
chrisandreae: Since we're making the parent cache generic and application owned, let's call it `parent_cache_group`.

PR #110, opened 2018-10-03
Record tree changes

chrisandreae:
Adds recursive tracking of changes to owned tree children during deserialization, to facilitate hooks that clear caches only on change.

Breaking changes:
* Use internal tracking for attributes and newness instead of delegating to model: clients need to call `x_changed!` in custom deserialize, and `model_is_new!` in custom resolve. This means that VM::AR attribute changes will now correctly correspond to viewmodel attributes instead of peeking through to the model.

* Fully visit and save node and all children before calling on_change hook: this means that hooks that inspect the model's changes will need to inspect `previous_changes`

* `before_save` hook is removed, replaced with `before_validate`. `before_validate` is also called from VM::Record.

### DIFF
--- a/iknow_view_models.gemspec
+++ b/iknow_view_models.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "iknow_cache"
   spec.add_dependency "iknow_params"
   spec.add_dependency "safe_values"
+  spec.add_dependency "keyword_builder"
 
   spec.add_dependency "concurrent-ruby"
   spec.add_dependency "jbuilder"
--- a/iknow_view_models.gemspec
+++ b/iknow_view_models.gemspec
@@ -44,5 +44,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "pg", '~> 0.18' # As of 5.1.4, Rails runtime check excludes pg 1.x, see #31669
   spec.add_development_dependency "pry"
   spec.add_development_dependency "rake"
+  spec.add_development_dependency "rspec-expectations"
   spec.add_development_dependency "sqlite3"
 end
--- a/lib/view_model/active_record/update_operation.rb
+++ b/lib/view_model/active_record/update_operation.rb
@@ -108,40 +108,22 @@ class ViewModel::ActiveRecord
             debug "-> #{debug_name}: Updating points-to association '#{reflection.name}'"
 
             association = model.association(reflection.name)
-            child_model =
+            new_target =
               if child_operation
                 child_ctx = viewmodel.context_for_child(association_data.association_name, context: deserialize_context)
-                child_operation.run!(deserialize_context: child_ctx).model
-              else
-                nil
+                child_viewmodel = child_operation.run!(deserialize_context: child_ctx)
+                if !association_data.shared? && child_viewmodel.previous_changes.changed_tree?
+                  viewmodel.children_changed!
+                end
+                child_viewmodel.model
               end
-            association.replace(child_model)
+            association.replace(new_target)
             debug "<- #{debug_name}: Updated points-to association '#{reflection.name}'"
           end
 
-          # Placing the edit check here allows it to consider the previous and
-          # current state of the model before it is saved. For example, but
-          # comparing #foo, #foo_was, #new_record?. Note that edit checks for
-          # deletes are handled elsewhere.
-
-          changes = viewmodel.changes
-          if changes.new? || changes.changed_attributes.present? || changes.changed_associations.present?
-            debug "-> #{debug_name}: Validating changes"
-            # invoke pre-save hook
-            viewmodel.before_save(changes, deserialize_context: deserialize_context)
-
-            # Invoke any validation hooks before attempting to save.
-            viewmodel.validate!
-
-            # The hooks before this might have caused additional changes. All
-            # changes should be checked by the policy, so we have to recalculate
-            # the change set. Note that changes to associated children that point
-            # to this parent are recorded in `changes`, but their models have not
-            # yet been updated.
-            final_changes = viewmodel.changes
-            deserialize_context.run_callback(ViewModel::Callbacks::Hook::OnChange, viewmodel, changes: final_changes)
-            hook_control.record_changes(final_changes)
-          end
+          # validate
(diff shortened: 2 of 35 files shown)


lib/view_model/utils/collections.rb
```diff
@@ -1,21 +1,27 @@
...
   module Collections
     def self.count_by(enumerable)
-      enumerable.each_with_object(Hash.new(0)) do |el, counts|
-        key         = yield(el)
-        counts[key] += 1 unless key.nil?
+      enumerable.each_with_object({}) do |el, counts|
+        key = yield(el)
+
+        unless key.nil?
+          counts[key] ||= 0
```
thefloweringash: What's the motivation to rewrite this? Avoiding absent keys from having "0" value?

While you're here, the following is clearer, imo.
```ruby
counts[key] = (counts[key] || 0) + 1
```

lib/view_model/utils/collections.rb
```diff
@@ -1,21 +1,27 @@
...
+        end
       end
     end
 
     refine Array do
       def contains_exactly?(other)
-        vals = to_set
-        other.each { |o| return false unless vals.delete?(o) }
-        vals.blank?
+        mine   = count_by { |x| x }
```
thefloweringash: What's the motivation to rewrite this?

lib/view_model/active_record/update_operation.rb
```diff
@@ -175,21 +150,30 @@ def run!(deserialize_context:)
...
+                  child_viewmodel.model
+                end
               end
 
             association.target = new_target
 
             debug "<- #{debug_name}: Updated pointed-to association '#{reflection.name}'"
           end
 
+          # Save and clear the changes
```
thefloweringash: Misleading comment.

lib/view_model/utils/collections.rb
chrisandreae: It was incorrect when there were multiple instances of the same value.

test/unit/view_model/access_control_test.rb
```diff
@@ -437,116 +440,311 @@ def test_ancestry
...
+        m = new_model_with_child.tap(&:save!)
+
+        vm, ctx = alter_by_view!(viewmodel_class, m) do |view, _refs|
+          view['child']['name'] = 'c'
+        end
+
+        # The parent node itself wasn't changed, so must not have been
+        # valid_edit checked
+        refute(ctx.was_edited?(vm.to_reference))
+        assert(vm.previous_changes.changed_children?)
```
thefloweringash: Consider making this an `asset_changes_match`, for completeness and for the reader to keep up their rhythm.

test/unit/view_model/access_control_test.rb
```diff
@@ -437,116 +440,311 @@ def test_ancestry
...
+        assert_changes_match(c_changes, true, false, false, ['name'], [])
+      end
+
+      it 'records an edited child without a tree change' do
+        vm, ctx = alter_by_view!(viewmodel_class, create_model!) do |_view, refs|
+          refs.values.first.merge!('name' => 'b')
+        end
+
+        refute(ctx.was_edited?(vm.to_reference))
+        refute(vm.previous_changes.changed_children?)
```
thefloweringash: Consider `assert_changes_match` here too.

test/unit/view_model/access_control_test.rb
```diff
@@ -437,116 +440,311 @@ def test_ancestry
...
+      it 'records a deleted child' do
+        vm, ctx = alter_by_view!(viewmodel_class, create_model!) do |view, refs|
+          view['child'] = nil
+          refs.clear
+        end
 
-      l2_edits = edit_check(ctx, ViewModel::Reference.new(ListView, l2.id))
-      assert_equal([[], [], true], l2_edits)
+        vm_changes = ctx.valid_edit_changes(vm.to_reference)
+        assert_changes_match(vm_changes, false, false, false, [], ['child'])
```
thefloweringash: `refute(ctx.was_edited?(old_child.to_reference))` ?

test/unit/view_model/access_control_test.rb
```diff
@@ -437,116 +440,311 @@ def test_ancestry
...
+          view['children'] << { '_ref' => 'new1' }
+          refs['new1'] = { '_type' => child_view_name, 'name' => 'b' }
+        end
+
+        vm_changes = ctx.valid_edit_changes(vm.to_reference)
+        assert_changes_match(vm_changes, false, false, false, [], ['children'])
+
+        new_child = vm.children.detect { |c| c.name == 'b' }
+        c_changes = ctx.valid_edit_changes(new_child.to_reference)
+        assert_changes_match(c_changes, true, false, false, ['name'], [])
```
thefloweringash: Does not test that preserved children weren't edited. The rest of the tests do.

iknow_view_models.gemspec
```diff
@@ -43,5 +44,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "pg", '~> 0.18' # As of 5.1.4, Rails runtime check excludes pg 1.x, see #31669
   spec.add_development_dependency "pry"
   spec.add_development_dependency "rake"
+  spec.add_development_dependency "rspec-expectations"
```
thefloweringash: Did you land up using these anywhere?

test/unit/view_model/active_record/has_many_test.rb
```diff
@@ -231,19 +230,19 @@ def test_replace_associated_has_many_functional
...
+      update([{ '_type' => 'Child', 'id' => old_children.first.id, 'name' => 'renamed p1c1' }])
     end
 
     nc = pv.replace_associated(:children, update, deserialize_context: context)
+    new_child = nc.detect { |c| c.name == 'new_child' }
 
-
-    expected_edit_checks = [ViewModel::Reference.new(ParentView, @parent1.id),
-                            ViewModel::Reference.new(ChildView,  nil),
+    expected_edit_checks = [pv.to_reference,
```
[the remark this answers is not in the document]
chrisandreae: I tried to change all the references constructed from ViewModels to use `to_reference`, but didn't change anything else. I could refactor the remainder of the tests to wrap any bare models in viewmodels?

PR #112, opened 2018-11-13
Parse PostgreSQL unique constraint violation messages

chrisandreae:
Make a best effort to extract the constraint name and the violated column and include in metadata

### DIFF
--- a/lib/view_model/deserialization_error.rb
+++ b/lib/view_model/deserialization_error.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class ViewModel
   class DeserializationError < ViewModel::AbstractErrorWithBlame
     status 500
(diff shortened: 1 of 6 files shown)


lib/view_model/deserialization_error.rb
```diff
@@ -306,6 +308,99 @@ def initialize(detail, nodes = [])
...
+          # value list.
+          return nil unless stream.delete_prefix!(DETAIL_INFIX)
+
+          [identifiers, stream]
+        end
+
+        private
+
+        QUOTED_IDENTIFIER   = /\A"(?:[^"]|"")+"/
+        UNQUOTED_IDENTIFIER = /\A(?:\p{Alpha}|_)(?:\p{Alnum}|_)*/
```
thefloweringash: `/\A(?:\p{Alpha}|_)(?:\p{Alnum}|_)*/` looks like it simplifies to `/\A(?:\p{Alpha}|_)+/`

lib/view_model/deserialization_error.rb
```diff
@@ -306,6 +308,99 @@ def initialize(detail, nodes = [])
...
+        end
+
+        self.new(err.message, constraint, columns, values, nodes)
+      end
+
+      class << self
+        DETAIL_PREFIX = 'Key ('
+        DETAIL_SUFFIX = ') already exists.'
+        DETAIL_INFIX  = ')=('
+        def parse_message_detail(detail)
```
thefloweringash: This reads fine, seems sensible, but contains a decent amount of logic not covered by tests.

lib/view_model/deserialization_error.rb
```diff
@@ -306,6 +308,99 @@ def initialize(detail, nodes = [])
...
+          stream = detail.dup
+
+          return nil unless stream.delete_prefix!(DETAIL_PREFIX)
+          return nil unless stream.delete_suffix!(DETAIL_SUFFIX)
+
+          # The message should start with an identifier list: pop off identifier
+          # tokens while we can.
+          identifiers = []
+
+          identifier = parse_identifier(stream)
```
thefloweringash: Awkward to have to unroll the loop body once due to the ordering of conditions. Would you prefer `loop/break`?

lib/view_model/deserialization_error.rb
chrisandreae: Alnum vs Alpha

PR #115, opened 2019-03-13
Provide default deserialization for primitive viewmodels

chrisandreae:
Similar to default serialization.

### DIFF
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -57,12 +57,15 @@ class ViewModel
       attrs.each { |attr| attribute(attr, **args) }
     end
 
-    def attribute(attr, **args)
+    def attribute(attr, **_args)
       unless attr.is_a?(Symbol)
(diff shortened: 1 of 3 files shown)


lib/view_model.rb
```diff
@@ -148,9 +155,43 @@ def serialize_to_hash(viewmodel, serialize_context: new_serialize_context)
...
+                                  view_hash[attr],
+                                  references: references,
+                                  deserialize_context: deserialize_context)
+          end
+        end
+
+        deserialize_context.run_callback(ViewModel::Callbacks::Hook::BeforeValidate, viewmodel)
+        viewmodel.validate!
+
+        # More complex viewmodels can use track changes to persistent backing
```
thefloweringash: Missing word?

chrisandreae: Yup

PR #117, opened 2019-04-03
Configure for CircleCI, test Rails 6 beta

chrisandreae:


.circleci/config.yml
```diff
@@ -0,0 +1,60 @@
...
+          command: bundle -v
+
+      - run:
+          name: Create gem cache key
+          command: cat Gemfile ./gemfiles/*.gemfile > /tmp/gemfiles
+
+      # Restore bundle cache
+      - restore_cache:
+          keys:
+            - iknow_viewmodels-{{ checksum "/tmp/gemfiles" }}
```
thefloweringash: Looking at the build output:
```
Skipping cache generation, cache already exists for key: iknow_viewmodels-ZnQkOstAa8cZw6P+KZJlZPbJf5MCj5j8EgLyS_aW_4s=
Found one created at 2019-04-03 08:08:35 +0000 UTC
```

You might want to cache based on lock files instead of Gemfiles, since this looks like it will drift over time and not be updated. Lockfile generation is an unavoidable part of this CI workflow, so there shouldn't be any additional time cost, and makes the caching almost (or entirely?) pure.

chrisandreae: Yup, but it was awkward.

PR #131, opened 2019-09-02
[WIP] prepend for functional collection updates

chrisandreae:
So this is a minimal and backward-compatible change, but I don't think I like the API much. Help me bike-shed it?

### DIFF
--- a/lib/view_model/active_record.rb
+++ b/lib/view_model/active_record.rb
@@ -18,6 +18,7 @@ class ViewModel::ActiveRecord < ViewModel::Record
   VALUES_ATTRIBUTE       = "values"
   BEFORE_ATTRIBUTE       = "before"
   AFTER_ATTRIBUTE        = "after"
+  PREPEND_ATTRIBUTE      = "prepend"
 
   require 'view_model/utils/collections'
   require 'view_model/active_record/association_data'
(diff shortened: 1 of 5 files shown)


chrisandreae:
Bikeshed?

thefloweringash:
> backward-compatible change

I'm not yet seeing the part that makes this backward-compatible.

With the `before` and `after` expression I assumed we had a way to express exhaustive regions as a sort of gap-lock. We don't, and I'm not convinced the additional complexity would have any benefit to consumers.

The factoring of `before` and `after` into `prepend` + `anchor` does give us the desired prepend and decreases the set of invalid requests. Which part of this don't you like?

chrisandreae:
It's a long time ago, hard to recover exactly what I was thinking.

> >  backward-compatible change

> I'm not yet seeing the part that makes this backward-compatible.

I'm pretty sure the comment referred to the first commit, not the second "alternative" commit

> The factoring of before and after into prepend + anchor does give us the desired prepend and decreases the set of invalid requests. Which part of this don't you like?

I think that's exactly what I liked about prepend and anchor tool. What I didn't like as much was it wasn't backwards compatible.

thefloweringash:
> I'm pretty sure the comment referred to the first commit, not the second "alternative" commit

Oh, oops. I just read the diff.

When using before/after the consistent approach is to introduce a handle for the start and end of the collection, probably in the form of a sentinel value. This means the prepend is expressed as `before: $list_start`. Previously we let append be the implicit default, but this would also be generalized to `after: $list_end`, and would leave the door open for talking about ranges like you can with actual elements.

I remain unconvinced that this level of expressiveness is a good idea. Could we go with the breaking change with a migration from the (currently) less expressive before/after? Just to ease deployments with existing clients.

chrisandreae:
I think I'm OK with that. Hopefully the migration in the fupdate code won't be too brittle.

PR #136, opened 2020-03-16
Sort root update operations for more consistent lock order on writes

chrisandreae:


lib/view_model/active_record/update_context.rb
```diff
@@ -94,11 +94,16 @@ def build_root_update_operations(root_updates, referenced_updates)
...
+      # Combine our root and referenced updates, and separate by viewmodel type.
+      # Sort by id where possible to obtain as close to possible a deterministic
+      # update order to avoid database write deadlocks. This can't be entirely
+      # comprehensive, since we can't control the order that shared references
+      # are referred to from roots (and therefore visited).
       updates_by_viewmodel_class =
         root_updates.lazily
           .map { |root_update| [nil, root_update] }
           .concat(referenced_updates)
+          .sort_by { |_, update_data| update_data.metadata.id.to_s }
```
thefloweringash: The `.to_s` here is unfortunate when a project is using entirely integer keys, but I don't have a solution.

chrisandreae: Note that this also handles the case of `nil` id (for new records).

PR #144, opened 2020-09-30
Add ViewModel migrations

chrisandreae:
`ViewModel::Record`s may now use `migrates` blocks to define a JSON-level transformation of the serialized data to previous `schema_version`s. VM::AR::Controller and VM::Cache are augmented to use these migrations to support requesting and supplying views at specific versions.

### DIFF
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -11,8 +11,9 @@ class ViewModel
   TYPE_ATTRIBUTE      = "_type"
   VERSION_ATTRIBUTE   = "_version"
   NEW_ATTRIBUTE       = "_new"
+  MIGRATED_ATTRIBUTE  = "_migrated"
 
-  Metadata = Struct.new(:id, :view_name, :schema_version, :new) do
+  Metadata = Struct.new(:id, :view_name, :schema_version, :new, :migrated) do
     alias :new? :new
   end
 
@@ -99,8 +100,9 @@ class ViewModel
       type_name      = hash.delete(ViewModel::TYPE_ATTRIBUTE)
       schema_version = hash.delete(ViewModel::VERSION_ATTRIBUTE)
       new            = hash.delete(ViewModel::NEW_ATTRIBUTE) { false }
+      migrated       = hash.delete(ViewModel::MIGRATED_ATTRIBUTE) { false }
 
-      Metadata.new(id, type_name, schema_version, new)
+      Metadata.new(id, type_name, schema_version, new, migrated)
     end
 
     def extract_reference_only_metadata(hash)
@@ -108,7 +110,7 @@ class ViewModel
       id             = hash.delete(ViewModel::ID_ATTRIBUTE)
       type_name      = hash.delete(ViewModel::TYPE_ATTRIBUTE)
 
-      Metadata.new(id, type_name, nil, false)
+      Metadata.new(id, type_name, nil, false, false)
     end
 
     def extract_reference_metadata(hash)
@@ -232,6 +234,11 @@ class ViewModel
 
     def accepts_schema_version?(schema_version)
       schema_version == self.schema_version
+    end
+
+    def schema_hash(schema_versions)
+      version_string = schema_versions.to_a.sort.join(',')
+      Base64.urlsafe_encode64(Digest::MD5.digest(version_string))
     end
 
     def preload_for_serialization(viewmodels, serialize_context: new_serialize_context, include_referenced: true, lock: nil)
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -11,6 +11,10 @@ class ViewModel
   TYPE_ATTRIBUTE      = "_type"
   VERSION_ATTRIBUTE   = "_version"
   NEW_ATTRIBUTE       = "_new"
+
+  # Migrations leave a metadata attribute _migrated on any views that they
+  # alter. This attribute is accessible as metadata when deserializing migrated
+  # input, and is included in the output serialization sent to clients.
   MIGRATED_ATTRIBUTE  = "_migrated"
 
   Metadata = Struct.new(:id, :view_name, :schema_version, :new, :migrated) do
--- a/lib/view_model/active_record/controller.rb
+++ b/lib/view_model/active_record/controller.rb
@@ -62,12 +62,64 @@ module ViewModel::ActiveRecord::Controller
   end
 
   included do
-    etag { self.viewmodel_class.deep_schema_version }
+    etag { migrated_deep_schema_version }
+  end
+
+  def parse_viewmodel_updates
+    super.tap do |update_hash, refs|
+      if migration_versions.present?
+        migrator = ViewModel::UpMigrator.new(migration_versions)
+        migrator.migrate!([update_hash, refs], references: refs)
+      end
+    end
+  end
+
+  def prerender_viewmodel(*)
+    super do |jbuilder|
+      yield(jbuilder) if block_given?
+
+      # migrate the resulting structure before it's serialized to a json string
+      if migration_versions.present?
+        tree = jbuilder.attributes!
+        migrator = ViewModel::DownMigrator.new(migration_versions)
+        migrator.migrate!(tree, references: tree['references'])
+      end
+    end
   end
 
   private
 
   def viewmodel_id
     parse_param(:id)
+  end
+
+  def migration_versions
+    @migration_versions ||=
+      begin
+        versions = parse_param(
+          :versions,
+          default: {},
+          with: IknowParams::Serializer::HashOf.new(
+            IknowParams::Serializer::String, IknowParams::Serializer::Integer))
+
+        migration_versions = {}
+
+        versions.each do |view_name, required_version|
+          viewmodel_class = ViewModel::Registry.for_view_name(view_name)
+
+          if viewmodel_class.schema_version != required_version
+            migration_versions[viewmodel_class] = required_version
+          end
+        rescue ViewModel::DeserializationError::UnknownView
+          # Ignore requests to migrate types that no longer exist
+          next
+        end
+
+        migration_versions.freeze
+      end
+  end
+
+  def migrated_deep_schema_version
+    ViewModel::Migrator.migrated_deep_schema_version(viewmodel_class, migration_versions, include_referenced: true)
   end
 end
--- a/lib/view_model/record.rb
+++ b/lib/view_model/record.rb
@@ -10,6 +10,7 @@ class ViewModel::Record < ViewModel
   attr_accessor :model
 
   require 'view_model/record/attribute_data'
+  require 'view_model/migratable_view'
 
   class << self
     attr_reader :_members
@@ -25,6 +26,9 @@ class ViewModel::Record < ViewModel
       @_members       = {}
       @abstract_class = false
       @unregistered   = false
+
+      include ViewModel::MigratableView
+      initialize_as_migratable_view
 
       @generated_accessor_module = Module.new
       include @generated_accessor_module
@@ -207,7 +211,7 @@ class ViewModel::Record < ViewModel
   end
 
   def serialize_view(json, serialize_context: self.class.new_serialize_context)
-    json.set!(ViewModel::ID_ATTRIBUTE, model.id) if model.respond_to?(:id)
+    json.set!(ViewModel::ID_ATTRIBUTE, self.id) if stable_id?
     json.set!(ViewModel::TYPE_ATTRIBUTE, self.view_name)
     json.set!(ViewModel::VERSION_ATTRIBUTE, self.class.schema_version)
 
--- a/lib/view_model/record.rb
+++ b/lib/view_model/record.rb
@@ -12,6 +12,8 @@ class ViewModel::Record < ViewModel
   require 'view_model/record/attribute_data'
   require 'view_model/migratable_view'
 
+  include ViewModel::MigratableView
+
   class << self
     attr_reader :_members
     attr_accessor :abstract_class, :unregistered
@@ -26,9 +28,6 @@ class ViewModel::Record < ViewModel
       @_members       = {}
       @abstract_class = false
       @unregistered   = false
-
-      include ViewModel::MigratableView
-      initialize_as_migratable_view
 
       @generated_accessor_module = Module.new
       include @generated_accessor_module
--- a/lib/view_model/test_helpers/arvm_builder.rb
+++ b/lib/view_model/test_helpers/arvm_builder.rb
@@ -4,7 +4,8 @@ class ViewModel::TestHelpers::ARVMBuilder
   # Building an ARVM requires three blocks, to define schema, model and
   # viewmodel. Support providing these either in an spec argument or as a
   # dsl-style builder.
-  Spec = Struct.new(:schema, :model, :viewmodel) do
+  Spec = Struct.new(:schema, :model, :viewmodel)
+  class Spec
     def initialize(schema:, model:, viewmodel:)
       super(schema, model, viewmodel)
     end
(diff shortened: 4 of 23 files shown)


lib/view_model/active_record/controller.rb
```diff
@@ -62,12 +62,64 @@ def destroy(serialize_context: new_serialize_context, deserialize_context: new_d
...
+      if migration_versions.present?
+        migrator = ViewModel::UpMigrator.new(migration_versions)
+        migrator.migrate!([update_hash, refs], references: refs)
+      end
+    end
+  end
+
+  def prerender_viewmodel(*)
+    super do |jbuilder|
+      yield(jbuilder) if block_given?
```
kevingriffin, line 79: Consider `defined?(yield)` here for performance, as this is likely a pretty frequented path.

https://medium.com/rubycademy/the-defined-keyword-in-ruby-b7a5a5a48e1e

chrisandreae: 1: Wow, that's amazing, Ruby never fails to surprise with new lows in its implementation. That `defined?`, a method-like thing that interrogates the current binding is different from `block_given?`, a method-like thing that interrogates the current binding is quite a 🦜 .

2: It's not a particularly hot path, since it's only called once per controller method; I think I'd rather go with the more idiomatic version to be less surprising.

kevingriffin: Agreed that it isn't needed, but I think the difference is that `block_given?` is a method, and `defined?` is a keyword.

thefloweringash (commented):
Implementation seems fine (some minor comments), but I want to talk about the conceptual models for migrations.

- One is the rails way that has a line of migrations. There's no pathfinding, everything proceeds in a single direction, and sometimes the line is cut by migrations that are one way. When a one way migration is hit there's no choice but to stop.
- Another way is a full graph of mappings between versions. Each operation runs along the shortest path between a target version and the source version. Paths may be overlapping and asymmetric. This reminds me a lot of the problem of packet routing.

What's implemented is the middle of the two options. The full graph exists, but isn't considered for pathfinding. Instead paths are required to be symmetrical, while allowing (and expecting) overlapping paths. If overlapping paths weren't expected, there wouldn't be a graph library dependency.

If the full graph was used, there would be some surprising paths where a single operation may include up and down migrations. As a small win you'd be able to remove `OneWayError`, since you'd know if you had a path before starting. I don't think that the complexity of the full graph is warranted. I'm wondering if the middle ground is.

lib/view_model/test_helpers/arvm_builder.rb
```diff
@@ -4,7 +4,8 @@ class ViewModel::TestHelpers::ARVMBuilder
   # Building an ARVM requires three blocks, to define schema, model and
   # viewmodel. Support providing these either in an spec argument or as a
   # dsl-style builder.
-  Spec = Struct.new(:schema, :model, :viewmodel) do
+  Spec = Struct.new(:schema, :model, :viewmodel)
```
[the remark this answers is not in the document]
chrisandreae: Something quite irritating: without it, -Wall used by the tests spits the completely spurious error
```
/Users/chris/workspace/iknow_view_models/lib/view_model/test_helpers/arvm_builder.rb:17: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/chris/workspace/iknow_view_models/lib/view_model/test_helpers/arvm_builder.rb:8: warning: The called method `initialize' is defined here
```

lib/view_model/record.rb
```diff
@@ -26,6 +27,9 @@ def initialize_as_viewmodel_record
       @abstract_class = false
       @unregistered   = false
 
+      include ViewModel::MigratableView
+      initialize_as_migratable_view
```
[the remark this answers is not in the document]
chrisandreae: I started with that: the problem with that is that `include` is a noop if it's included by an ancestor, resulting in `ApplicationView` being a migratable view, but none of its decendents.

lib/view_model.rb
```diff
@@ -99,16 +100,17 @@ def extract_viewmodel_metadata(hash)
       type_name      = hash.delete(ViewModel::TYPE_ATTRIBUTE)
       schema_version = hash.delete(ViewModel::VERSION_ATTRIBUTE)
       new            = hash.delete(ViewModel::NEW_ATTRIBUTE) { false }
+      migrated       = hash.delete(ViewModel::MIGRATED_ATTRIBUTE) { false }
```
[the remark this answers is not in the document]
chrisandreae: That's not entirely true: raw data without `_migrated` set comes in from the user. It gets migrated, resulting in a tree of hashes with `_migrated`. Then that tree gets deserialized, which involves parsing it with `extract_viewmodel_metadata`. Like other metadata, the `_migrated` is parsed out by that.

test/unit/view_model/active_record/migration_test.rb
```diff
@@ -0,0 +1,150 @@
...
+      )
+    end
+
+    it 'migrates' do
+      migrate!
+
+      assert_equal(expected_result, subject)
+    end
+
+    describe 'with a version not in the specification' do
```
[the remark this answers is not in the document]
chrisandreae: Migrations are applied according to the request by the client. If the client says in its request that `versions = { "Blah": 
6 }`, it's saying "when sending you data, I will send you Blah@6 and expect it to be migrated up to the current version, and when you send me data I expect you to migrate it down to the current version".

If the client says Blah@6 and then proceeds to supply a serialized blah at a different version, well, that's not the version that it asked for migrations to be performed on: it's left alone, and it'll fall through to result in an incorrect version error when deserialized.

lib/view_model.rb
```diff
@@ -11,8 +11,9 @@ class ViewModel
   TYPE_ATTRIBUTE      = "_type"
   VERSION_ATTRIBUTE   = "_version"
   NEW_ATTRIBUTE       = "_new"
+  MIGRATED_ATTRIBUTE  = "_migrated"
 
-  Metadata = Struct.new(:id, :view_name, :schema_version, :new) do
+  Metadata = Struct.new(:id, :view_name, :schema_version, :new, :migrated) do
```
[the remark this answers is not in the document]
chrisandreae: What would you like to see? How about 
> Migrations leave a metadata attribute `_migrated` on any views that they alter. This attribute is accessible as metadata when deserializing migrated input, and is included in the output serialization sent to clients.

chrisandreae:
We discussed the migration tree in detail out of band. The options amount to:

1 (current): migrations are an undirected graph, you simply find the shortest path and use it in both directions. That migration path may or may not be able to be run in both directions. It doesn’t matter if there’s a longer path that could run in both directions, you don’t use it, you must use the same path in both directions or not at all, and it must be the shortest one. The migration between 1 and 4 is (1<->4), regardless of whether one of those directions throws.

2 (possible): migrations define a pair of graphs, for the up and down migrations. A graph only has an edge for a migration if the migration can run in that direction. You separately find the shortest path of up migrations and the shortest path of down migrations. This may be asymmetric. The migration up from 1 to 4 is 1<->4, the migration down from 4 to 1 is 4<->3 + 3<->1

3 (we know we don’t want this): migrations define a single directed graph made up of all the up and down edges, we just try and find the shortest path even if that path uses both ups and downs.

2 is more powerful, but asymmetric routes might prove to be surprising in practise. We also couldn't think of any practical example where asymmetic routes would be needed: if we have symmetric 1-2 and 2-3, and we write a 1-3, there is nothing that would make one of those directions impossible.

Because it's easier to relax later than restrict, we're going to go with the current more constrained model until we have a reason to relax it.

We also talked about whether it was useful to make it an error to write an unreachable migration. For example, if we have 1-2, 2-3 and 1-3, the 1-2 is never used: migrations are always done to or from the latest version, and we can reach all earlier versions without using it. This seems like a piece of beneficial linting, but not actual ill-defined code.

lib/view_model/record.rb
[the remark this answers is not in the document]
thefloweringash: You can added to the `inherited` method in `ClassMethods`:

```diff
diff --git a/lib/view_model/migratable_view.rb b/lib/view_model/migratable_view.rb
index 44ddb4d..8e0631d 100644
--- a/lib/view_model/migratable_view.rb
+++ b/lib/view_model/migratable_view.rb
@@ -10,6 +10,11 @@ module ViewModel::MigratableView
   extend ActiveSupport::Concern
 
   class_methods do
+    def inherited(base)
+      super
+      base.initialize_as_migratable_view
+    end
+
     def initialize_as_migratable_view
       @migrations_lock   = Monitor.new
       @migration_classes = {}
diff --git a/lib/view_model/record.rb b/lib/view_model/record.rb
index fd8d908..546043d 100644
--- a/lib/view_model/record.rb
+++ b/lib/view_model/record.rb
@@ -12,6 +12,8 @@ class ViewModel::Record < ViewModel
   require 'view_model/record/attribute_data'
   require 'view_model/migratable_view'
 
+  include ViewModel::MigratableView
+
   class << self
     attr_reader :_members
     attr_accessor :abstract_class, :unregistered
@@ -27,9 +29,6 @@ class ViewModel::Record < ViewModel
       @abstract_class = false
       @unregistered   = false
 
-      include ViewModel::MigratableView
-      initialize_as_migratable_view
-
       @generated_accessor_module = Module.new
       include @generated_accessor_module
     end
`

File: lib/view_model.rb (hunk 6 above)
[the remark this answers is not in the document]
Comment (thefloweringash) [reply lib/view_model.rb:16]: Seems reasonable to me.

File: lib/view_model.rb (hunk 4 above)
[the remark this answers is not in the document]
Comment (thefloweringash) [reply lib/view_model.rb:103]: At first glance it's not obvious if the user is allowed to set this. I think it's covered with the above comment though.

**Comment by thefloweringash** [comment]
I don't see the linking `.rubocop.yml` file in the commit introducing the linter setup:

```yaml
inherit_gem:
  rubocop-iknow: rubocop.yml
```

lib/view_model/record.rb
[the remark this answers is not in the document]
chrisandreae: oh nice, I didn't think of that

test/unit/view_model/active_record/migration_test.rb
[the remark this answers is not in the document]
chrisandreae: It'd also be legitimate for the migrator to say "Hey, you promised me version 6" and blow up early on seeing any other version. Do you think that'd be better?

lib/view_model.rb
```diff
@@ -234,6 +240,11 @@ def accepts_schema_version?(schema_version)
       schema_version == self.schema_version
     end
 
+    def schema_hash(schema_versions)
+      version_string = schema_versions.to_a.sort.join(',')
```
butchler, line 244: @chrisandreae I'm trying to figure out how to calculate the hash in JS, but I'm not sure how Ruby's `sort` method deals with the array of tuples returned by `schema_versions`. I think I figured out how it works based on the documentation here: https://docs.ruby-lang.org/en/2.0.0/Array.html

> Each object in each array is compared (using the <=> operator).
>
> Arrays are compared in an “element-wise” manner; the first two elements that are not equal will determine the return value for the whole comparison.

I assume `<=>` uses standard comparisons for strings and numbers, so I can probably write a custom comparator that handles comparing tuples in this way.

Is this correct?

Alternatively, do you think it would make sense to first join the pairs before sorting? i.e.:

```
# Not sure if this is valid Ruby
version_string = schema_versions.to_a.map({ |pair| pair.join(',') }).sort.join(',')
```

That way the JS implementation could do the exact same thing (i.e. `Object.entries(schemaVersions).map(pair => pair.join(',')).sort().join(',')`).

butchler: Now that I think about it, maybe these two sorting methods will always produce the same result? They seem more or less equivalent, but I'm not sure how to prove whether or not they are exactly equivalent.

chrisandreae:
```
irb(main):007:0> { 'd' => 1, 'a' => 3, 'b' => 4 }.to_a
=> [["d", 1], ["a", 3], ["b", 4]]
irb(main):008:0> _.sort
=> [["a", 3], ["b", 4], ["d", 1]]
irb(main):009:0> _.join(',')
=> "a,3,b,4,d,1"
```

The Array [comparator](https://docs.ruby-lang.org/en/2.0.0/Array.html#method-i-3C-3D-3E) sorts element-wise: comparing the first element, and if they're equivalent then the second element and so on.

But yes, as you update, the source of the tuples is a map, and so the first element is structurally guaranteed to be unique: so the sort order is necessarily defined by purely by the view name.

butchler: I don't seem to be getting the same hashes as what's in `migration_version_hash.json`. For example, if I put the `Brand` version 3 info in `irb`:

```
irb(main):040:1* {
irb(main):041:1*   "Brand": 3,
irb(main):042:1*   "BrandAuthProvider": 1,
irb(main):043:1*   "BrandCanonicalLanguage": 1,
irb(main):044:1*   "BrandDomain": 2,
irb(main):045:1*   "BrandExternalSite": 2,
irb(main):046:1*   "BrandLogo": 1,
irb(main):047:1*   "Curriculum": 1,
irb(main):048:1*   "Organization": 5,
irb(main):049:1*   "OrganizationLicensePartner": 1,
irb(main):050:1*   "OrganizationSiteCustomization": 1,
irb(main):051:1*   "Text": 1,
irb(main):052:1*   "TextTranslation": 1
irb(main):053:0> }.to_a.sort.join(',')
=> "Brand,3,BrandAuthProvider,1,BrandCanonicalLanguage,1,BrandDomain,2,BrandExternalSite,2,BrandLogo,1,Curriculum,1,Orga
nization,5,OrganizationLicensePartner,1,OrganizationSiteCustomization,1,Text,1,TextTranslation,1"
irb(main):054:0> Digest::MD5.digest(_)
=> "k3@M/\x89l\xCA#\xF2e\xBBO-O\xF9"
irb(main):055:0> Base64.urlsafe_encode64(_)
=> "azNATS-JbMoj8mW7Ty1P-Q=="
irb(main):056:0>
```

I get `azNATS-JbMoj8mW7Ty1P-Q==` as the version hash, but that data is under the key `Uzs_1YH6Cob8W-Snv0kcfA` in `migration_version_cache.json`.

Am I doing something wrong?

chrisandreae: > `Digest::MD5.digest(_)`

is the problem. We previously used md5 to calculate the digest, but Gordon pointed out that browsers don't include a md5 algorithm by default, so that would unnecessarily increase the bundle size. So the hash is now calculated by using the first 16 bytes of the SHA256, which the browser natively supports.

The hash is calculated from the versions map by `ViewModel.schema_hash` in the `iknow_viewmodels_gem` here:
https://github.com/iknow/iknow_view_models/blob/master/lib/view_model.rb#L249

butchler: Oh I see. I just followed the link here from the backend PR, so I didn't realize there were further changes.

Interesting, I didn't know the browser supports SHA256 natively.

chrisandreae: Here's where Gordon linked me to when discussing it: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

butchler: I think I've gotten the SHA-256 based hash calculation working now, but the data doesn't seem to be migrated when using `versions_hash`. For example, this request does not require migrated data: `/api/brands?domain=localhost.devdomain.name&versions_hash=Uzs_1YH6Cob8W-Snv0kcfA`. However, it works when using the non-hash version params (e.g. `versions[Brand]=3&...`).

@chrisandreae Are you able to reproduce this?

butchler: Oh, I think I found the problem: The backend checks for the existence of `versions_hash`, but then actually uses `version_hash`.

PR #156, opened 2021-01-26
gemspec: update email

kevingriffin:
Our organization email isn't edge@iknow.jp. In fact, that's an invalid email.


chrisandreae (approved):
Seems good.

PR #160, opened 2021-02-19
Ruby 3.0 support

thefloweringash:
This also drops support for ruby 2.6.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ executors:
     parameters:
       ruby-version:
         type: string
-        default: "2.6"
+        default: "2.7"
       pg-version:
         type: string
         default: "11"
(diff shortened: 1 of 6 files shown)


chrisandreae:
> This also drops support for ruby 2.6.

but not from the circleci config?

thefloweringash:
> > This also drops support for ruby 2.6.
> 
> but not from the circleci config?

We have a choice: drop 2.6 and use the nice `...` syntax, or keep 2.6 and use whatever syntax is required to do so. Is 2.6 important anywhere?

chrisandreae:
Does Engoo use viewmodels? That might be a reason, if they're not yet on new Rails.

But that's not really what I meant: 
Screen Shot 2021-02-19 at 9 00 04 PM
I just meant that this branch doesn't drop ruby 2.6 from the circleci test file.

thefloweringash:
> I just meant that this branch doesn't drop ruby 2.6 from the circleci test file.

I was keeping this as an explicit failure to make it extremely obvious that ruby 2.6 support is going away. Updated to be bold and fully drop ruby 2.6 and older.

chrisandreae:
Does this deserve a minor version bump to convey the changed requirements?

PR #162, opened 2021-03-24
Allow rendering using cache even when the roots aren't cacheable

chrisandreae:
Light refactor of ViewModel::AR::Cache to move rendering of roots into the CacheWorker, and to change the entrypoint to a class method to allow rendering from non-cacheable roots.

### DIFF
--- a/lib/view_model.rb
+++ b/lib/view_model.rb
@@ -173,6 +173,17 @@ class ViewModel
       Jbuilder.new { |json| serialize(viewmodel, json, serialize_context: serialize_context) }.attributes!
     end
 
+    def serialize_from_cache(views, migration_versions: {}, locked: false, serialize_context:)
+      plural = views.is_a?(Array)
+      views = Array.wrap(views)
+
+      json_views, json_refs = ViewModel::ActiveRecord::Cache.render_viewmodels_from_cache(
+                    views, locked: locked, migration_versions: migration_versions, serialize_context: serialize_context)
+
+      json_views = json_views.first unless plural
+      return json_views, json_refs
+    end
+
     def encode_json(value)
       # Jbuilder#encode no longer uses MultiJson, but instead calls `.to_json`. In
       # the context of ActiveSupport, we don't want this, because AS replaces the
--- a/lib/view_model/active_record/cache.rb
+++ b/lib/view_model/active_record/cache.rb
@@ -11,6 +11,30 @@ class ViewModel::ActiveRecord::Cache
 
   attr_reader :viewmodel_class
 
+  class << self
+    def render_viewmodels_from_cache(viewmodels, migration_versions: {}, locked: false, serialize_context: nil)
+      if viewmodels.empty?
+        return [], {}
+      end
+
+      ids = viewmodels.map(&:id)
+      # ideally the roots wouldn't have to all be the same type
+      viewmodel_class = viewmodels.first.class
+      serialize_context ||= viewmodel_class.new_serialize_context
+
+      render_from_cache(viewmodel_class, ids,
+                        initial_viewmodels: viewmodels,
+                        migration_versions: migration_versions,
+                        locked: locked,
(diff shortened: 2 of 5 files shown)


lib/view_model/active_record/cache.rb
```diff
@@ -35,85 +53,98 @@ def clear
     @cache_group.invalidate_cache_group
   end
 
+  # @deprecated, replaced by class methods
```
thefloweringash, line 56: I don't think yard tags use commas, see [examples](https://rubydoc.info/gems/yard/file/docs/Tags.md#deprecated).

lib/view_model.rb
```diff
@@ -173,6 +173,14 @@ def serialize_to_hash(viewmodel, serialize_context: new_serialize_context)
       Jbuilder.new { |json| serialize(viewmodel, json, serialize_context: serialize_context) }.attributes!
     end
 
+    def serialize_from_cache(views, migration_versions: {}, locked: false, serialize_context:)
+      plural = views.is_a?(Array)
+      views = Array.wrap(views)
+      json_views, json_refs = ViewModel::ActiveRecord::Cache.render_viewmodels_from_cache(views, locked: locked, migration_versions: migration_versions, serialize_context: serialize_context)
```
thefloweringash, line 179: Super long line (even though it's just refactored).

lib/view_model/active_record/cache.rb
```diff
@@ -35,85 +53,98 @@ def clear
...
+
+          # If initial root viewmodels were provided, call hooks on any
+          # viewmodels which were rendered from the cache to ensure that the
+          # root is visible (in isolation). Other than this, no traversal
+          # callbacks are performed for cache-rendered views. This particularly
+          # requires care for references: if a visible view may refer to
+          # non-visible cacheable views, those referenced views will not be
+          # access control checked.
+          initial_viewmodels&.each do |v|
+            next unless cached_serializations.has_key?(v.id)
```
thefloweringash, line 116: It's been a while since I looked at this, but this looks like quite a departure from the current implementation. Justify why the presence of a cached serialization is sufficient to skip the access control check? In the previous implementation they're always checked, which effectively means "you can use their cached views if you can see the root", so access controls _always_ apply. In the new implementation the meaning is "you can use cached views if they have been cached", which sometimes skips access controls.

However I believe it's correct to only check cacheable views, since non-cached views will always be checked as part of rendering.

lib/view_model/active_record/cache.rb
```diff
@@ -35,85 +53,98 @@ def clear
...
 
-      # Resolve references
-      worker.resolve_references!
+        # Render remaining views. If initial viewmodels have been locked, we may
+        # use them to serialize from, otherwise we must reload with share lock
+        # in find_and_preload.
+        available_viewmodels =
+          if locked
+            initial_viewmodels&.each_with_object({}) do |vm, h|
+              h[vm.id] = vm if missing_ids.include?(vm.id)
```
thefloweringash, line 128: I don't see where `missing_ids` is defined. It looks like it should be using `ids_to_render`. Missing test case?

lib/view_model.rb
chrisandreae: Worth fixing

lib/view_model/active_record/cache.rb
chrisandreae: Yeah. Previously the callbacks were double-invoked in the case of a cache miss, which seemed undesirable.

lib/view_model/active_record/cache.rb
chrisandreae: Yeah, that's a missed rename from the refactor, and absolutely a missing test case.

chrisandreae: fixed and added test cases

thefloweringash (approved):
LGTM

PR #163, opened 2021-04-28
Nested controller method locking

chrisandreae:
Support locking owner viewmodel in nested controller methods. Before this, there was no way for a controller extending this to add a lock that would be within the transaction before the (de)serialization operation, without also capturing the `render_json_string`.


thefloweringash (approved):
LGTM

PR #164, opened 2021-05-19
Remove context from eager includes

chrisandreae:
Before the prune/include support was dropped, `eager_includes` required a `serialize_context` to calculate pruning. This wasn't removed from the signature at the time pruning was removed, even though it became no longer used. This had consequent effects on the signatures of methods that called eager_includes, such as `ViewModel.preload_for_serialization`, `VM::AR.load`, and `VM::AR.find`.


thefloweringash (approved):
LGTM

PR #166, opened 2021-06-21
Support STI polymorphism in collection associations

chrisandreae:
Previously we assumed a collection association had a single `viewmodel_class`: we could use it to construct a view for a given model, and check it for list membership. With STI polymorphism, it may have multiple viewmodel_classes. We must resolve the viewmodel independently per model, and the list membership must be consistent among the viewmodel_classes. Extend AssociationData to lookup and validate this association ordering.

This does not yet support polymorphism in external AssociationManipulation, but that's consistent with its current behaviour of not supporting through polymorphism.

Additionally including minor fixes to tests and migrations.

### DIFF
--- a/lib/view_model/active_record/association_data.rb
+++ b/lib/view_model/active_record/association_data.rb
@@ -196,6 +196,21 @@ class ViewModel::ActiveRecord::AssociationData
     viewmodel_classes.first
   end
 
+  def ordered?
+    @ordered ||=
+      if through?
+        direct_viewmodel._list_member?
+      else
+        list_members = viewmodel_classes.map { |c| c._list_member? }.uniq
+
+        if list_members.size > 1
+          raise ArgumentError.new('Inconsistent associated views: mixed list membership')
+        end
+
+        list_members[0]
+      end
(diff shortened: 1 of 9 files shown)


lib/view_model/active_record/association_data.rb
```diff
@@ -196,6 +196,21 @@ def viewmodel_class
     viewmodel_classes.first
   end
 
+  def ordered?
```
thefloweringash, line 199: There are a few remaining places that use `._list_member?` instead of `ordered?`. It'd be nice to clean those up, even if there's no observable difference (`load_associated`, `append_associated`).

thefloweringash (approved):
LGTM

lib/view_model/active_record/association_data.rb
chrisandreae: Sure, done. There's still one remaining use of _list_member?, in a context that doesn't have the association data: https://github.com/iknow/iknow_view_models/blob/support-sti/lib/view_model/active_record/update_operation.rb#L652

PR #168, opened 2021-06-23
Polymorphic association data

chrisandreae:
Properly mark association_data to STI tables as polymorphic: this is required for example to properly form `eager_includes` preload specs


thefloweringash (approved):
LGTM

PR #172, opened 2021-10-07
Migrator: skip directly to values in fupdates

chrisandreae:
Within functional updates, `_type` is used with a context-aware meaning that doesn't indicate an actual viewmodel of that type: in particular, the values for the `delete` action and the before/after of the append action specify the `_type` of the collection type, but in fact must only be _type/id, not full viewmodels, and must not be migrated.

Handle this structure in the up migrator.

### DIFF
--- a/lib/view_model/migrator.rb
+++ b/lib/view_model/migrator.rb
@@ -75,6 +75,32 @@ class ViewModel
   class UpMigrator < Migrator
     private
 
+    def migrate_tree!(node, references:)
+      if node.is_a?(Hash) && node[ViewModel::TYPE_ATTRIBUTE] == ViewModel::ActiveRecord::FUNCTIONAL_UPDATE_TYPE
+        migrate_functional_update!(node, references: references)
+      else
+        super
+      end
+    end
+
+    NESTED_FUPDATE_TYPES = ['append', 'update'].freeze
+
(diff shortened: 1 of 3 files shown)


lib/view_model/migrator.rb
```diff
@@ -75,6 +75,34 @@ def migrate_viewmodel!(_view_name, _version, _view_hash, _references)
...
+    NESTED_FUPDATE_TYPES = ['append', 'update'].freeze
+
+    # The functional update structure uses `_type` internally with a
+    # context-dependent meaning. Retrospectively this was a poor choice, but we
+    # need to account for it here.
+    def migrate_functional_update!(node, references:)
+      actions = node[ViewModel::ActiveRecord::ACTIONS_ATTRIBUTE]
+      return unless actions
+
+      actions.each do |action|
```
thefloweringash, line 95: Consistency nitpick: here you use a guard for nil actions, below you use `values&.each` for absent values. Is this intentional?

thefloweringash (approved):
LGTM

lib/view_model/migrator.rb
chrisandreae: No, I changed one and not the other, I'll make it consistent.

PR #174, opened 2021-10-28
Allow inheriting a migration from another viewmodel class

chrisandreae:
Useful in the case that multiple views are defined on the same underlying data, that underwent the same transformation.


thefloweringash (approved):
LGTM

PR #175, opened 2021-10-28
deserialization: GC removed nodes marked dependent: :delete_all

chrisandreae:
This case was missed, causing removed :delete_all children to silently be not deleted.
- [x] Bump version


lib/view_model/active_record/update_context.rb
```diff
@@ -14,7 +14,7 @@ def initialize(*)
       def release!
         model = viewmodel.model
         case association_data.direct_reflection.options[:dependent]
-        when :delete
+        when :delete, :delete_all
           model.delete
         when :destroy
```
thefloweringash, line 19: While you're here, does this also need `:destroy_all`?

thefloweringash (commented):
Seems reasonable.

lib/view_model/active_record/update_context.rb
chrisandreae: Is there a `destroy_all`?

thefloweringash: It seemed consistent, but it doesn't seem to match reality.

```ruby
# belongs_to
    def self.valid_dependent_options
      [:destroy, :delete, :destroy_async]
    end

# has_one
    def self.valid_dependent_options
      [:destroy, :destroy_async, :delete, :nullify, :restrict_with_error, :restrict_with_exception]
    end

# has_many
    def self.valid_dependent_options
      [:destroy, :delete_all, :nullify, :restrict_with_error, :restrict_with_exception, :destroy_async]
    end
```

So we should probably add `:destroy_async`.

chrisandreae: yeah, though we wouldn't (and wouldn't want to) do it asynchronously (?)

PR #176, opened 2021-12-15
Record `changed?` when asserting a mutable value on creation

chrisandreae:
Even if it's the model's default value. This allows callbacks and access control to differentiate between an explicitly asserted value and a fallback to the model default.

Also refactors the dummy model in `RecordTest`: previously the "model default values" were anything but that, and couldn't differentiate between a value that was asserted and that was not.

Bumps semantic minor version, as this slightly changes the contract with library users.

### DIFF
--- a/test/unit/view_model/record_test.rb
+++ b/test/unit/view_model/record_test.rb
@@ -57,11 +57,30 @@ class ViewModel::RecordTest < ActiveSupport::TestCase
     let(:model_body)     { nil }
     let(:viewmodel_body) { nil }
 
+    # Generate an ActiveModel-like keyword argument constructor.
+    def generate_model_constructor(model_class, model_defaults)
+      args = model_class.members
+      params = args.map do |arg_name|
+        "#{arg_name}: self.class.__constructor_default(:#{arg_name})"
+      end
+
+      <<-SRC
+        def initialize(#{params.join(", ")})
+          super(#{args.join(", ")})
+        end
+      SRC
+    end
+
     let(:model_class) do
       mb = model_body
-      Struct.new(*attributes.keys) do
-        class_eval(&mb) if mb
-      end
+      mds = model_defaults
+
+      model = Struct.new(*attributes.keys)
+      constructor = generate_model_constructor(model, mds)
+      model.class_eval(constructor)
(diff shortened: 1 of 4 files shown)


test/unit/view_model/record_test.rb
```diff
@@ -285,31 +323,47 @@ def validate!
...
         end
-        assert_equal('read_only', ex.attribute)
       end
 
-      it 'rejects update if changed' do
-        new_view = default_view.merge('read_only' => 'written')
-        ex = assert_raises(ViewModel::DeserializationError::ReadOnlyAttribute) do
-          viewmodel_class.deserialize_from_view(new_view, deserialize_context: update_context)
+      describe 'attempting a change' do
+        let(:default_view_values) { { read_only: 'attempted change' } }
```
thefloweringash, line 352: It's been quite a while since I read these tests, but I think expressing it like this is making it harder to reason about.

IIUC, the`default_view` is a "neutral" serialization that is to be used as a base for the tests to extend with their exceptional cases. To make an attempted change, as also seen in the previous version of the test, you add your change to the neutral view: `new_view = default_view.merge(read_only: 'attempted_change')`.

In this case where the value is used twice, I'd prefer a `let(:new_view) { ... }`.

test/unit/view_model/record_test.rb
```diff
@@ -338,10 +392,25 @@ def new_record?
       end
     end
 
+    describe 'with unspecified attributes falling back to the model default' do
+      let(:attributes)           { { value: {} } }
+      let(:default_view_values)  { { } }
```
thefloweringash, line 397: To be consistent with my understanding of `default_view_values`, this should be expressed as `let(:default_values) { { value: 5 } }`, and the deserialize should use `deserialize_from_view({}, ...)`

thefloweringash (approved):
New functionality LGTM, some quibbles about the test stucture.

test/unit/view_model/record_test.rb
chrisandreae: I agree, and have changed in the new refactor

test/unit/view_model/record_test.rb
chrisandreae: I addressed this slightly differently, by separating the test subjects from the defaults.

PR #178, opened 2022-02-16
VM::AccessControl::Composed: allow checkers to return a custom error

chrisandreae:
A custom error is signaled by returning `failure(AnErrorObject)` from a checker body.

In the case of `unless` checkers, this error is guaranteed to be raised.
In the case of `if` checkers, the first custom error to be returned (if any) will be raised if all `if` checkers fail.


thefloweringash (approved):
LGTM

PR #180, opened 2022-05-31
AccessControl::Tree: fix inheritance of root children checks

chrisandreae:
Additionally, improve `inspect` for composed nodes.

### DIFF
--- a/lib/view_model/access_control/tree.rb
+++ b/lib/view_model/access_control/tree.rb
@@ -192,8 +192,6 @@ class ViewModel::AccessControl::Tree < ViewModel::AccessControl
       def inspect_checks
         checks = super
         if root?
-          checks << 'no root checks'
-        else
           checks << "root_children_visible_if: #{root_children_visible_ifs.map(&:reason)}"            if root_children_visible_ifs.present?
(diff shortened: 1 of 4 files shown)


lib/view_model/access_control/tree.rb
```diff
@@ -192,8 +192,6 @@ def root?
       def inspect_checks
         checks = super
         if root?
-          checks << 'no root checks'
```
chrisandreae, line 195: this was just completely backwards - on a root note it output a lie, and on a non-root node it fruitlessly went through all the checkers that would necessarily be empty

lib/view_model/access_control/tree.rb
```diff
@@ -252,16 +250,16 @@ def validate_root!(view, context)
 
     def save_root_visibility!(traversal_env)
       result = check_delegates(traversal_env,
-                               self.class.each_check(:root_children_visible_ifs,      ->(a) { a.is_a?(Node) }),
-                               self.class.each_check(:root_children_visible_unlesses, ->(a) { a.is_a?(Node) }))
+                               self.class.each_check(:root_children_visible_ifs,      ->(a) { a < Node }),
```
chrisandreae, line 253: The `inherit_from` checker is the class, not the instance!

thefloweringash (approved):
LGTM

Issue #165, opened 2021-06-04
Generalize `through` associations


chrisandreae:
Currently, `through:` association are implemented with the assumption that they're for use with many-to-many associations, which don't make sense at all to be owned. Because of that, a `through:` association to a non-root isn't supported. This is a bit over-strict, since there are other valid cases to have an association use a hidden join table, such as an ordered polymorphic collection.

Achieving this will require a hefty refactor of `UpdateOperation#build_updates_for_collection_referenced_association` to generalize it to work for both referenced and nested targets.

An additional feature that we may want to consider supporting should we refactor this is to support polymorphic associations where the targets may be either (referenced) root types or (nested) owned types. This would allow us to better represent the common vs custom text associations as a single polymorphic association.

PR #182, opened 2022-08-09
[WIP] Support renaming viewmodel classes in migrations

chrisandreae:


lib/view_model/migratable_view.rb
```diff
@@ -19,12 +19,14 @@ def initialize_as_migratable_view
       @migrations_lock   = Monitor.new
       @migration_classes = {}
       @migration_paths   = {}
-      @realized_migration_paths = true
+      @previous_names    = {}
+      @realized_paths    = true
+      @previous_name_cache = nil
```
thefloweringash, line 24: This has unfortunate overlap
 - `previous_names` = points at which the name changed, very internal
 - `previous_names_cache` = all valid names, aka "versioned view names"
```suggestion
      @versioned_view_names = nil
```

And similarly change `cache_previous_names!` to `cache_versioned_view_names!`.

test/unit/view_model/active_record/migration_test.rb
```diff
@@ -273,6 +273,84 @@ def new_model
...
+      let(:subject) { v1_serialization.deep_dup }
+
+      let(:expected_result) do
+        current_serialization.deep_merge(
+          {
+            'data' => {
+              ViewModel::MIGRATED_ATTRIBUTE => true,
+            },
+          },
+        )
```
thefloweringash, line 344: Nitpick: formatted differently to the down migration.
```suggestion
        current_serialization.deep_merge({ 'data' => { ViewModel::MIGRATED_ATTRIBUTE => true } })
```

thefloweringash (commented):
LGTM, small number of comments.

lib/view_model/migrator.rb
```diff
@@ -5,16 +5,16 @@ class Migrator
...
-      def migrated_deep_schema_version(viewmodel_class, required_versions, include_referenced: true)
+      def migrated_deep_schema_version(viewmodel_class, client_versions, include_referenced: true)
         deep_schema_version = viewmodel_class.deep_schema_version(include_referenced: include_referenced)
 
-        if required_versions.present?
+        if client_versions.present?
           deep_schema_version = deep_schema_version.dup
 
-          required_versions.each do |required_vm_class, required_version|
+          client_versions.each do |required_vm_class, client_version|
```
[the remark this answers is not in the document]
chrisandreae: I buy that argument. It's not quite as bad as suggested, since the old and new versions were both used in the cache keys so it already was being invalidated no matter what, but I think it just is less surprising to have the name-at-the-time in the structure produced here.

lib/view_model/registry.rb
```diff
@@ -75,10 +93,37 @@ def clear_removed_classes!
...
+        if vm.respond_to?(:versioned_view_names)
+          vm.versioned_view_names
+        else
+          { vm.schema_version => vm.view_name }
+        end
+
+      versioned_view_names.each do |previous_version, previous_name|
+        versions_for_name = (@viewmodel_classes_by_name_and_version[previous_name] ||= {})
+
+        if (prev_vm = versions_for_name[previous_version]) && prev_vm != vm
```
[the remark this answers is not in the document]
chrisandreae: > Is it expected that a viewmodel will register itself multiple times

I saw this happening in tests

> Similarly, should we be checking that viewmodel_classes_by_name is unique? Admittedly, it's a lot harder to accidentally have two classes register the same name.

Yeah, that makes sense, I've added that.

lib/view_model/migratable_view.rb
chrisandreae: Makes sense, I'll do that.

chrisandreae:
So there are two issues that are blocking this:

1: registering the historical names and versions in the registry at `register`/`resolve_deferred_classes` time is insufficient. The deferral is meant to be for long enough to a class to set the `view_name`. It's quite common within the body of a viewmodel class to end up hitting `for_view_name` and therefore resolving the deferred classes, including itself. If this happens before the `migrates` blocks, it'll be registered without any of its historical versions.

A possible solution to this would be to split the deferral in two, to resolve the class name and the historical versions independently. Then, `migrates` could validate that the latter resolution hasn't happened, effectively banning lookup of old versions (that would cause resolution) during the body of the class.

2: View aliases. Presently, alias names aren't registered at old versions at all, which is simply insufficient. But also, if a view has aliases, we don't know when they're valid from. We might have to extend the `renamed_from` to have multiple histories, and support something like `renamed from: 'ParagraphView', to: 'NewParagraphView'` to rename a non-default name. Then the old versions could be computed for all the alias names.

Alternatively, the alias names only exist for the MTI subclassing, which we don't like and wish we didn't have any more. If we were to implement #165, we could go ahead and release https://github.com/iknow/eikaiwa_content/pull/1598 which would remove the need for aliases, and we could drop that feature entirely.

chrisandreae:
Alternative approach to (1) that would require less careful maintenance of the deferred structures:

Rather than the registry maintaining the full map of name -> version -> class, it could instead have
* a map of current-canonical-name -> class
* a map of historical-name -> [classes], sorted by descending highest version

At construction time you need to make sure there is no version overlap.
Then when resolving at a certain version, you take the first class that agrees that it responds to that version.
When resolving at no-version-specified, you only take the current canonical name.


Then because not all the version resolution is done eagerly, the only “deferred” step has to be adding the main class name, and that only needs to be deferred until the class name actually exists.

Alias names and historical names can be added immediately: at the time a view_alias call or a migrates { renamed_from: } is encountered.

PR #184, opened 2022-10-24
Migration references

thefloweringash:


chrisandreae (approved):
Seems reasonable. Clearing a provided `references` gave me pause for a moment, but I can't actually see any reason that it's not OK.

PR #186, opened 2022-11-17
Fix migrations adding references

chrisandreae:
Migrations adding references could result in concurrent modification errors if a reference was already being visited. Support this properly, and ensure that the thus-added references are themselves visited for migration.


thefloweringash (approved):
LGTM

PR #188, opened 2023-02-01
Add configuration for deferred constraints via deserialize context

chrisandreae:
Usually, we want to validate deferred database constraints during a deserialization operation to catch and wrap any errors that we caused. Occasionally, if the deserialization is not the entire operation, we want to defer this validation until after all changes have been completed. Add a flag to DeserializeContext::SharedContext to control automatic validation, enabled by default.


thefloweringash (approved):
LGTM

PR #189, opened 2023-02-16
add github actions

chrisandreae:

PR #190, opened 2023-05-12
[v2.x] Fix argument forwarding

thefloweringash:
In particular new_serialize_context should forward keyword arguments instead of relying on `*args` to handle keyword arguments. While we're here though, use `...` where possible, taking hints from the mainline branch.

---

~We don't have any CI at the moment. See #189, and the back port to `v2.x` #191.~


chrisandreae (approved):
All looks reasonable to me

PR #192, opened 2023-05-23
gemspec: relax version range of iknow_params

thefloweringash:
It seems like different bundler versions find different solutions. Some use iknow_view_models 2.8.4 with iknow_params 2.3.0, and some use iknow_view_models 2.10.0 with iknow_params 2.2.0. It's probably some heuristic about which package is best to be out of date. Let's just avoid all this confusion by using the latest iknow_params.


chrisandreae:
I don't know of a reason that this wouldn't work.

PR #193, opened 2023-05-30
VM::Controller: add hook to render metadata in the same context

chrisandreae:
This would permit metadata serialization to contribute referenced viewmodels.


chrisandreae:
Oh, one awkward thing here is that this wouldn't place nicely with the `prerender_json_view` used to render cached content, because in that the references are already preserialized as json and there's no serialize context to contribute more to. That alone is probably a blocker for this idea, unless you've got any insight.

PR #198, opened 2024-01-15
VM::AR::Controller: introduce specified_migration_versions

thefloweringash:
Returns all migration versions specified by the client, for consumers that want to use this information directly.


chrisandreae (approved):
Looks right; don't forget to bump the gem version

PR #199, opened 2024-01-24
Allow expressing create/update to the anonymous child of a singular association

chrisandreae:
This requires two extensions, to be able to express both "update existing" or "create-or-update".

We do this with extensions to the interpretation of `_new`:
  * `true`  - Always create a new model, using the id if provided, error if          a model with that id already exists
  * `false` - Never create a new model. **If id isn’t provided, it's an error            unless the viewmodel is being deserialized in the context of a            singular association, in which case it's implicitly referring            to the current child of that association, and is an error if            that child doesn't exist.**
 * `nil`    - Create a new model if `id` is not specified, otherwise update the            existing record with `id`, and error if it doesn't exist.
 * `'auto'` - **Can only be used when the viewmodel is being deserialized in the            context of a singular association. `id` must not be specified. If            the association currently has a child, update it, otherwise            create a new one.**

### DIFF
--- a/lib/iknow_view_models/version.rb
+++ b/lib/iknow_view_models/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module IknowViewModels
-  VERSION = '3.7.7'
+  VERSION = '3.8.0'
 end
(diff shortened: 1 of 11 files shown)


thefloweringash (approved):
LGTM

chrisandreae:
Oh, I was too eager and hit the merge button before squashing the `fixup!`. Too late now, it's already gone to rubygems, sorry.

PR #200, opened 2024-02-14
cache delete: permissive calling convention

chrisandreae:
Mimic convenience overloads by runtime type-checking the arguments, to accept ~either an array or~ splat of ids or model/viewmodel objects.

### DIFF
--- a/lib/view_model/active_record/cache.rb
+++ b/lib/view_model/active_record/cache.rb
@@ -49,8 +49,12 @@ class ViewModel::ActiveRecord::Cache
     @migrated_cache = @migrated_cache_group.register_cache(cache_name)
   end
 
+  # Handles being called with either ids or model/viewmodel objects
   def delete(*ids)
     ids.each do |id|
+      id = id.id if id.respond_to?(:id)
+      raise ArgumentError.new unless id.is_a?(Numeric) || id.is_a?(String)
+
       @cache_group.delete_all(@cache.key.new(id))
     end
   end
--- a/lib/iknow_view_models/version.rb
+++ b/lib/iknow_view_models/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module IknowViewModels
-  VERSION = '3.9.0'
+  VERSION = '3.9.1'
 end


chrisandreae:
I'm not entirely confident I like the array-or-splat dispatch, because it's not a very common pattern. The id-or-model dispatch is very Rails-y and extremely common.

thefloweringash:
> I'm not entirely confident I like the array-or-splat dispatch, because it's not a very common pattern. The id-or-model dispatch is very Rails-y and extremely common.

I'm mostly motivated by avoiding mistakes. We've found that passing models instead of id is a mistake we've made more than once. Accepting models as part of the interface removes that class of bug. We haven't found that passing arrays is a common mistake (but could be considered a natural extension of the api). So I'm happy to take the conservative approach.

Could we, in the name of identifying mistakes, check that the ids are the correct type and raise if they're not? I think the check would have to be approximately as follows. Sniffing types here does not fill me with joy.

```ruby
  def delete(*ids)
    ids.each do |id|
       id = id.id if id.respond_to?(:id)
       raise ArgumentError.new unless id.is_a?(Numeric) || id.is_a?(String)
       @cache_group.delete_all(@cache.key.new(id))
     end
  end
```

chrisandreae:
> Could we, in the name of identifying mistakes, check that the ids are the correct type and raise if they're not? I think the check would have to be approximately as follows. Sniffing types here does not fill me with joy.

I buy that. Silently failing is unacceptable, but we don't have to work to coerce anything plausible into the right shape. Also, if we go beyond Rails' usual conventions in anything-compatibility, we're laying a bit of a trap for users by conditioning them to expect that level of anything-compatibility when they're not going to get it from the rest of the system.

lib/view_model/active_record/cache.rb
```diff
@@ -49,8 +49,13 @@ def initialize(viewmodel_class, cache_group: nil)
     @migrated_cache = @migrated_cache_group.register_cache(cache_name)
   end
 
+  # Handles being called with either an array or a splat of
+  # ids or model/viewmodel objects
```
thefloweringash, line 53: Comment no longer matches behavior.

chrisandreae: fixed

thefloweringash (approved):
LGTM

PR #201, opened 2024-02-19
Deserialization: visit members in declaration order if possible

chrisandreae:
We require members to be visited in declaration order so that we can safely
write viewmodels for models whose members have dependencies on the values of
other members.

We're limited to deserializing points-to and pointed-from associations before
and after save respectively, but we can at least correctly interleave attributes
and points-to associations.


thefloweringash (approved):
LGTM

PR #203, opened 2024-04-23
Extract migration parsing to a controller concern

chrisandreae:
Controllers that don't need a full VM::AR::Controller may still need to support parsing migrations.


thefloweringash (approved):
LGTM

PR #204, opened 2024-05-11
VM::Cache: retry without using cache on stale cached references

chrisandreae:
If a cached value contains a reference to a entity that no longer exists, this currently results in a RecordNotFound error. Instead, we should be able to treat this as a signal that the cached value is out of date, and re-resolve it from the database.

### DIFF
--- a/lib/view_model/active_record/cache.rb
+++ b/lib/view_model/active_record/cache.rb
@@ -30,8 +30,21 @@ class ViewModel::ActiveRecord::Cache
     end
 
     def render_from_cache(viewmodel_class, ids, initial_viewmodels: nil, migration_versions: {}, locked: false, serialize_context: viewmodel_class.new_serialize_context)
-      worker = CacheWorker.new(migration_versions: migration_versions, serialize_context: serialize_context)
-      worker.render_from_cache(viewmodel_class, ids, initial_viewmodels: initial_viewmodels, locked: locked)
+      ignore_existing = false
+      begin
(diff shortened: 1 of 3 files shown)


chrisandreae:
On encountering a stale value, we give up entirely on the cache contents and re-render (and re-populate the cache) from the database. This helps us avoid the worst-case scenario of repeatedly failing and retrying when retrieving many entities all with the same stale reference.

thefloweringash (approved):
LGTM

lib/view_model/active_record/cache.rb
```diff
@@ -181,8 +201,15 @@ def resolve_references!
...
 
-        viewmodels = find_and_preload_viewmodels(viewmodel_class, required_entries.keys,
-                                                 available_viewmodels: available_viewmodels)
+        viewmodels =
+          begin
+            find_and_preload_viewmodels(viewmodel_class, required_entries.keys,
+                                        available_viewmodels: available_viewmodels)
+          rescue ViewModel::DeserializationError::NotFound => e
+            # We tried to resolve a stale cached reference: we need to retry
+            # without using the stale value.
```
chrisandreae, line 210:
```suggestion
            # We encountered a reference to an entity that does not exist.
            # If this reference was potentially found in cached data, it
            # could be stale: we can retry without using the cache.
            # If the reference was obtained directly, it indicates invalid
            # data such as an invalid foreign key, and we cannot recover.
```

PR #209, opened 2024-12-17
Deserialization includes

thefloweringash:
This enables unconditional eager loading during the deserialization phase, which can be used to ensure that models are loaded in access controls and hooks.


chrisandreae (approved):
This seems really sensible! 👍

PR #213, opened 2025-09-22
Support down-migrations that need to premigrate to intermediates

chrisandreae:
When a migration moves attributes from one view to another, it might be necessary to premigrate a referenced view to obtain those attributes in the correct shape. But the contemporary version that it needs to be premigrated to might not be the eventual destination of that view, which isn't compatible with the current down-migrator.

Extend the down-migrator to support completing the down-migration of a partially-migrated intermediate.


thefloweringash (approved):
LGTM

PR #214, opened 2026-02-06
Add support for strict migration validation

chrisandreae:
With strict migration validation, every visited view must have a migration version specified by the client.


lib/view_model/record.rb
```diff
@@ -239,6 +239,18 @@ def serialize_members(json, serialize_context:)
     end
   end
 
+  # On its own, a Record view has no preloadable dependencies. But if its
+  # attributes are serialized with viewmodels, those attributes might themselves
+  # have preloadable dependencies.
+  def preloadable_dependencies(include_referenced: true)
```
thatsmydoing: Is this changed related to the checks?

thatsmydoing (approved):

lib/view_model/record.rb
chrisandreae: Oh, it’s not, that wasn’t meant to be in this branch.

gemfiles/rails_6_1.gemfile
thefloweringash, line 1: These generated gemfiles are a realization of the contents of the `Appraisals` file. You should clean that up while you're here.

thefloweringash (approved):

gemfiles/rails_6_1.gemfile
chrisandreae: oh, gosh, I wonder when the last time the generation was run, it looks so completely different

